PR #115, opened 2014-03-05
Add validation hooks for assertion ID, time range, recipient, and destination

kwugirl:
Originally at https://github.com/onelogin/ruby-saml/pull/103 - original description copied below:

This pull request contains four additions:
1. A validation hook for the assertion ID. We found that we needed to ensure that we rejected reused assertion IDs so as to prevent replay attacks.
2. A validation hook for the time range conditions. We wanted to enforce sensible time range choices so that our customers wouldn't be unnecessarily weakening their security by making their assertions valid for long periods of time.
3. A validation hook for the recipient attribute. We wanted to protect our customers from attacks where an unauthorized user gains access by using a valid assertion for a different third-party Service Provider.
4. A validation hook for the destination attribute. Similar to #3, we wanted to protect our customers from attacks where an unauthorized user gains access by using a valid assertion for a different third-party Service Provider.

All these additions are covered with new tests as well.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -135,6 +135,7 @@ module OneLogin
         validate_structure(soft)      &&
         validate_response_state(soft) &&
         validate_conditions(soft)     &&
+        validate_new_assertion_id(soft) &&
         document.validate_document(get_fingerprint, soft) &&
         validate_success_status(soft)
       end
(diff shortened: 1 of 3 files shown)


Lordnibbler:
@kwugirl, @curious-attempt-bunny If you still want this PR can you please rebase off master and make the proposed refactor?  we've switched to minitest from test::unit

pitbulk:
I gonna close that PR, we created a new one: https://github.com/onelogin/ruby-saml/pull/197

PR #171, opened 2014-11-12
Fix inconsistent results with using regex matches in decode_raw_saml

subfusc:
After running SAML code in production against a Norwegian IdP in a similar library
, it was found by Jørgen Fjeld <jorgen at veridit.no> that it is unreliable to
check for /^</. This is because the ^ matches any line and not just the first one.
It does not help replacing it with \A (Which is the beginning of the string) as
the zlib compressed data may start with a "<".

The Solution implemented in this patch is to check that the current string is Base64
complient. Because XML can't be Base64 compliant, we check if the string is
base64 encoded and if it isn't we return the string with the assumption that it
is the XML. If it is we try and decode + inflate. This make the function
idempotent and decode_raw_saml(decode_raw_saml(Response)) will yield the correct
result.

### DIFF
--- a/lib/onelogin/ruby-saml/saml_message.rb
+++ b/lib/onelogin/ruby-saml/saml_message.rb
@@ -37,16 +37,23 @@ module OneLogin
 
       private
 
+      ##
+      # Take a SAML object provided by +saml+, determine its status and return
+      # a decoded XML as a String.
+      #
+      # Since SAML decided to use the RFC1951 and therefor has no zlib markers,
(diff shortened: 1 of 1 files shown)


Lordnibbler:
@subfusc can you please rebase off `master` and remove the Nokogiri dependency change? we have moved to using Minitest and updated the dependency

subfusc:
@Lordnibbler cherry-picked the commit ontop of current master. So, done. ;)

PR #175, opened 2014-12-14
Introduce thread safety to SAML schema validation

phlipper:
Per the references below, use of `Dir.chdir` is not thread-safe. This usage was causing exceptions to be raised when running on Puma and in other multi-threaded environments.

This patch also moves the schema read up to a class instance, as this data is static and does not need to be read every time an assertion is validated. This boosts performance, especially in environments with higher throughput.

Thanks to @dannyb for the assistance!

References:

* https://www.ruby-forum.com/topic/165079
* https://bugs.ruby-lang.org/issues/9785
* http://www.justskins.com/forums/working-directory-in-thread-42304.html
* http://www.ruby-doc.org/core-2.1.5/Dir.html#method-c-chdir

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -151,19 +151,18 @@ module OneLogin
       end
 
       def validate_structure(soft = true)
-        Dir.chdir(File.expand_path(File.join(File.dirname(__FILE__), '..', '..', 'schemas'))) do
-          @schema = Nokogiri::XML::Schema(IO.read('saml-schema-protocol-2.0.xsd'))
-          @xml = Nokogiri::XML(self.document.to_s)
-        end
-        if soft
-          @schema.validate(@xml).map{
-            @errors << "Schema validation failed";
-            return false
-          }
-        else
-          @schema.validate(@xml).map{ |error| @errors << "#{error.message}\n\n#{@xml.to_s}";
-            validation_error("#{error.message}\n\n#{@xml.to_s}")
-          }
+        xml = Nokogiri::XML(self.document.to_s)
+
+        SamlMessage.schema.validate(xml).map do |error|
+          if soft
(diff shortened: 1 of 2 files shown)


cgthornt:
Wouldn't it make sense to not even change the directory and read the file relative to `__FILE__`? i.e.

```ruby
@schema ||= begin
  path = File.expand_path(File.join(File.dirname(__FILE__), '..', '..', 'schemas', 'saml-schema-protocol-2.0.xsd'))
  Nokogiri::XML::Schema(IO.read(path))
end
```

phlipper:
@cgthornt the `chdir` is required due to a known issue with Nokogiri and schema validation.

References:

* http://ktulu.com.ar/blog/2011/06/26/resolving-validation-errors-using-nokogiri-and-schemas/
* https://stackoverflow.com/questions/11996326/nokogirixmlschema-syntaxerror-on-schema-load
* https://github.com/dougalcorn/nokogiri-chdir-schema-validate-bug

Lordnibbler:
@phlipper any chance you can rebase off `master` and throw a test around the new method you introduced? `self.schema`

We updated the test suite to use Minitest and I just want to ensure its all passing  still before we merge.

Thank you for this improvement!

phlipper:
@Lordnibbler i have rebased on top of the latest master and tests are green.

Lordnibbler:
I'm :+1: on this but would love a simple unit test around this new `self.schema` method

phlipper:
@Lordnibbler what would you like to see for test coverage? I'm happy to add some, but there are no existing tests for the `SamlMessage` class at the moment. The subclasses are tested and passing so I felt confident in the change.

Lordnibbler:
@phlipper We will make a followup issue to test the `SamlMessage` class (unless you want to throw some tests around the class too).

Lordnibbler:
@cgthornt @pitbulk can you guys review?

PR #178, opened 2015-01-15
XMLSecurity::Document#sign_document refactor

luisvm:


luisvm:
@pitbulk shouldn't `XMLSecurity::Document#sign_document` return an instance of `XMLSecurity::SignedDocument` ?

pitbulk:
@luisvm 
Right now this method modify the root of the REXML::Document that call the SignedDocument method.

PR #179, opened 2015-01-16
Add support for setting the entity ID and name ID format when parsing metadata

jhdavids8:
Any reason we shouldn't do this?

Thanks for the gem, by the way! Great stuff.


carlosedf:
+1. Certificate would be useful too.

pitbulk:
:+1:

luisvm:
:+1:

PR #181, opened 2015-01-20
make schema validation threadsafe (eliminate runtime chdir)

draganm:
Eliminates runtime chdir when validating schemas


phlipper:
This is a duplicate of #175

PR #182, opened 2015-01-22
Signature algorithm would not be changed by signature method. Algorithm ...

polika78:
Signature algorithm would not be changed by signature method. Algorithm method's argument was not REXML:Element, so fixed it.

### DIFF
--- a/lib/xml_security.rb
+++ b/lib/xml_security.rb
@@ -102,7 +102,7 @@ module XMLSecurity
       signature_element = REXML::Element.new("ds:Signature").add_namespace('ds', DSIG)
       signed_info_element = signature_element.add_element("ds:SignedInfo")
       signed_info_element.add_element("ds:CanonicalizationMethod", {"Algorithm" => C14N})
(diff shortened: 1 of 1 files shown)


Lordnibbler:
@pitbulk @luisvm  thoughts on this? seems sane to me :+1:

pitbulk:
The algorithm method accepts  REXML::Element or an string that represent the signature algorithm like RSA_SHA1.  I don't see why is better pass the signed_method_element instead the signature_method parameter directly.

polika78:
The algorithm method is parsing the signature method as integer number (ex. sha256 -> 256) when the signature method is only REXML::Element. The signature method usually includes prefix like 'sha'. If this method accept both type Element or string for the signature method, I think that algorithm method should be fixed for parsing properly.

pitbulk:
Fixed at https://github.com/onelogin/ruby-saml/commit/e96becf57a725dd090aefff783c9fd37be900a47

PR #190, opened 2015-01-30
Adds Scoping element to auth request

tekt8tket:
## Summary
Adds a scoping element to the saml auth request
- adds 3 settings attributes to facilitate the data needed for building the XML
- mirrors SimpleSAMLphp functionality here: https://simplesamlphp.org/docs/1.7/simplesamlphp-scoping

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -111,6 +111,31 @@ module OneLogin
           end
         end
 
+        if settings.proxy_count || settings.idp_list || settings.requester_id
+          proxy_count_attribute = if settings.proxy_count
+            {
+              'ProxyCount' => settings.proxy_count.to_s
+            }
+          end
+
+          scoping = root.add_element 'samlp:Scoping', proxy_count_attribute
+
+          if settings.idp_list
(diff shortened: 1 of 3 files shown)


Lordnibbler:
@tekt8tket can you please rebase off master branch and update your tests to use minitest syntax?

pitbulk:
Hi @tekt8tket, sorry for the delay in the reply.

The ruby-saml toolkits support only set 1 IdP so the IdP scoping has not much sense.
We also try to keep the toolkit simple so we avoid to add functionality that is not required by the main customers of the toolkit.

robertpostill:
Hi @pitbulk I understand the desire to keep it simple but then that does beg the question of where a change like this should go?

I also would like the scoping functionality so I don't want to see the work @tekt8tket has done go to waste or land up in a fork.

pitbulk:
The unique solution I see is to create a "extended" branch and try to push there all the PRs that were rejected and try to keep it updated, but this only gonna be possible if we find collaborators.

Right now we maintaining several toolkits: php, java, python, dot-net and ruby. Also several plugins so you can underestand that the effort is huge. And I dont want to reduce time for maintaining the main ruby branch by maintaining that forked branch.

Issue #126, opened 2014-04-29
Take care of  XXE (Xml eXternal Entity) Attack


pitbulk:
XXE attacks: https://www.owasp.org/index.php/XML_External_Entity_%28XXE%29_Processing

The ruby-saml uses Nokogiri  at     lib/xml_security.rb , 74

Read about this Nokogiri vulnerability issue:
https://github.com/sparklemotion/nokogiri/issues/693

Issue #125, opened 2014-04-29
Limit entity expansion when parsing XML to avoid XEE attacks.


pitbulk:
XEE attacks details:  
https://www.ruby-lang.org/en/news/2008/08/23/dos-vulnerability-in-rexml/

Use entity_expansion_limit

https://www.ruby-lang.org/en/news/2013/02/22/rexml-dos-2013-02-22/

http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2014-8080

PR #192, opened 2015-02-04
Entity expand prevent

pitbulk:
Related to #125 and #126. Security issues: XXE Attacks


pitbulk:
@Lordnibbler What do you think about this PR?

Lordnibbler:
Is there a reason for setting `REXML::Document::entity_expansion_limit = 0` multiple times in different classes? If we always want this value to be 0, why not just set `REXML::Document::entity_expansion_limit = 0` just once, since it will be defined at the class level?

phlipper:
I agree with @Lordnibbler, otherwise LGTM

Lordnibbler:
Needs  rebase @pitbulk

PR #194, opened 2015-02-10
Relax nokogiri gem requirements

Lordnibbler:
## Status
**READY**

## Description
Adjust `nokogiri` gem requirement:
* Ruby 1.8.7 to use 1.5
* all newer versions of Ruby to use 1.6.x.

## Todos
- [x] dont use weird array syntax for dependency definitions
- [x] adjust gemspec

### DIFF
--- a/ruby-saml.gemspec
+++ b/ruby-saml.gemspec
@@ -25,28 +25,33 @@ Gem::Specification.new do |s|
   s.summary = %q{SAML Ruby Tookit}
   s.test_files = `git ls-files test/*`.split("\n")
 
-  s.add_runtime_dependency('uuid', ['~> 2.3'])
-  s.add_runtime_dependency('nokogiri', ['~> 1.5.0'])
+  s.add_runtime_dependency('uuid', '~> 2.3')
+  if RUBY_VERSION < '1.9'
+    # 1.8.7
+    s.add_runtime_dependency('nokogiri', '~> 1.5.10')
+  else
+    s.add_runtime_dependency('nokogiri', '~> 1.6.0')
+  end
 
-  s.add_development_dependency 'minitest',           ['~> 5.5']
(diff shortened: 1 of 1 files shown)


elskwid:
@Lordnibbler, good change for the version. That helps me in my case where the app is sitting at `1.6.5`. Thanks!

Lordnibbler:
@elskwid yes, sorry about this -- it was an overlooked change when we switched to Minitest.

ruby-saml.gemspec
```diff
@@ -25,28 +25,33 @@ Gem::Specification.new do |s|
...
   s.test_files = `git ls-files test/*`.split("\n")
 
-  s.add_runtime_dependency('uuid', ['~> 2.3'])
-  s.add_runtime_dependency('nokogiri', ['~> 1.5.0'])
+  s.add_runtime_dependency('uuid', '~> 2.3')
+  if RUBY_VERSION < '1.9'
+    # 1.8.7
+    s.add_runtime_dependency('nokogiri', '~> 1.5.10')
+  else
+    s.add_runtime_dependency('nokogiri', '~> 1.6')
```
phlipper: Shouldn't this be `~> 1.6.0`? Without using the _patch-level_ identifier, this runs the risk of v1.7+ breaking something in the future. Otherwise you could remove the conditional and just use `~> 1.5` and that should satisfy everything Ruby 1.8+.

williampratt:
:+1:

ruby-saml.gemspec
Lordnibbler: @phlipper I will make the change to `~> 1.6.0`, but unfortunately setting it to `~> 1.5` will still cause Ruby 1.8 to try to install Nokogiri 1.6.x, despite it not being compatible :(

PR #193, opened 2015-02-04
Decrypt support

pitbulk:
Initial decrypt support. Related to #9

Issue #186, opened 2015-01-29
Need to update README for encryption key settings


shyam-habarakada:
### This is regarding the current `decrypt` branch

I have gotten a version of what is the decrypt branch to work against ADFS 2.0 as an IdP. However, the process I had to go through to get it working was a lot of trial and error and reading through commits and source in that branch—The README did not help much.

This issue is to determine what to add to the README so I can send in a PR.

In my current setup, I am setting `settings.idp_cert`, `settings.certificate` and `settings.private_key`. These last two values are the public key I send to the IdP for encrypting the assertion in the `SAMLResponse` and my private key for decrypting the encrypted assertion, respectively.

The `settings.certificate` I assign however is exposed as a `signing` certificate in metadata XML. Shouldn't that be an `encryption` certificate instead?

In the tests that added around encryption, there is reference to both signing and encryption as well. See 86852424fcad8f9e22de1e0a1d534b4bdfc2fd22 

Shouldn't we support separate certificates for signing vs encryption?

@pitbulk since you are currently owning the `decrypt` branch, can you please comment?

Thanks

shyam-habarakada:
@buffym would love your input on this as well.

pitbulk:
Hi @shyam-habarakada,

I know that the standard let you use different private_key/cert for signing and encrypting.
My experience in SAML project says to me that 99% of people use the same thing for both method.

My main goal when implementing the OneLogin's SAML Toolkits was to keep the toolkit simple and cover all the main scenarios. We already have complex solutions with complex settings and I'm trying to avoid that philosophy.

I'm agree with you that when adding that decrypt method, I will need to update the Readme and reference that the key/cert will works for both sign and encrypt.

And maybe later, if someone that really need to use different keys/certs for that I can explain him how extend the toolkit for support that, but I want to avoid adding more complexity to the settings tp cover the 1% of the cases.

Thanks you for supporting the ruby-saml toolkit and be an active contributor ;)

buffym:
Due to time constraints on the project I am working on, I ended up doing encrypted assertions with a different gem. I was planning on taking a look back at this one once this branch was updated and merged. For what it's worth, I agree, w/ @pitbulk regarding signing/encryption keys; every case I've seen, they are the same.

shyam-habarakada:
I am seeing customers that have different certs, although that is not directly impacting us. As an SP, I can chose to use the same for both signing and encrypting.

[...]

Issue #172, opened 2014-11-13
Can ruby-saml deal with PingFederate's SSO ?


googya:
THIS IS NOT A ISSUE! 

Can ruby-saml deal with PingFederate's SSO ?  Or is there any method to sovle the problem?

buffym:
Sorry for the delayed response - no, I never got it working. Due to time constraints, I ended up using https://github.com/digidentity/libsaml for encrypted assertions. Still using ruby-saml for other parts of SSO in our app though.

googya:
@pitbulk 👍

kwent:
Same issue here getting invalid signature. What PingFederate is doing differently than other IdP provider ?

PR #196, opened 2015-02-19
Adding supported for Encrypted SAML 

ksinkar:
These changes are for adding support for **encrypted** SAML messages. [xmlenc] (https://github.com/digidentity/xmlenc) from has been used for encryption and decryption. All existing tests for ruby-saml are passing. I had to upgrade the timecop gem though, for the tests to pass on my local machine. Unfortunately these are failing on Travis CI for REE and 1.8.x 

@buffym @googya, could you please check and let me know if this works for your cases of Ping Federate and Encryption otherwise.

#9 #193 #186 #172

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -18,12 +18,16 @@ module OneLogin
       attr_reader :options
       attr_reader :response
       attr_reader :document
-
+      attr_reader :decoded_response
+      attr_reader :decrypted_response
+      
       def initialize(response, options = {})
         @errors = []
         raise ArgumentError.new("Response cannot be nil") if response.nil?
         @options  = options
-        @response = decode_raw_saml(response)
+        @decoded_response = decode_raw_saml(response)
+        @decrypted_response = decrypt_saml(@decoded_response, @options[:private_key_file_path])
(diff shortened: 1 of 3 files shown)


pitbulk:
Thanks @ksinkar, we will review all this code and integrate it in the ruby-saml gem this functionality for sure.

ksinkar:
@pitbulk, I had a look at the decrypt branch in which you are developing the decrypt functionality. It  seems that branch is only able to handle signing and checking of digest methods for encrypted assertions. I don't see any code for decrypting the encrypted assertions. 

Please correct me if I am wrong in the above inference.

I am eager to  get the decryption functionality into this gem. I suggest that you merge my changes into master or the decrypt branch as you see fit. As for the failing tests, it is  dependency problem of nokogiri which I have tried to solve by updating the xmlenc gemspec as well, but not yet been able to solve it. If you have any ideas please let me know and I shall solve them pronto.

pitbulk:
@ksinkar  Here is the code to decrypt assertions:
https://github.com/onelogin/ruby-saml/blob/decrypt/lib/onelogin/ruby-saml/response.rb#L258

pitbulk:
Replaced by #241. Right now we avoid the use of xmlenc.

PR #199, opened 2015-03-06
Change Nokogiri's runtime dependency to fix support for Ruby 1.8.7.

### LINKED ISSUE #198: 0.9.1 gem appears to require Nokogiri ~> 1.6.0 on Ruby 1.8.7
#194 loosened the runtime requirement on Nokogiri so that version `~> 1.6.0` would be accepted for Ruby 1.9 and above:

```ruby
if RUBY_VERSION < '1.9'
  # 1.8.7
  s.add_runtime_dependency('nokogiri', '~> 1.5.10')
else
  s.add_runtime_dependency('nokogiri', '~> 1.6.0')
end
```

However, this code does not appear to be working and `~> 1.6.0` is being enforced for 1.8.7 as well.

The easiest way to reproduce this issue is to create a clean project with the following Gemfile:

```ruby
source "http://rubygems.org"
gem "ruby-saml", "0.9.1"
```

This results in the following:

```
Fetching gem metadata from http://rubygems.org/..........
Resolving dependencies...
Installing systemu (2.6.4) 
Installing macaddr (1.7.1) 
Installing mini_portile (0.6.2) 
Installing nokogiri (1.6.6.2) 
Gem::InstallError: nokogiri requires Ruby version >= 1.9.2.
An error occurred while installing nokogiri (1.6.6.2), and Bundler cannot continue.
Make sure that `gem install nokogiri -v '1.6.6.2'` succeeds before bundling.
```

If you try to add a version constraint for Nokogiri in the Gemfile, you end up getting the following error instead:

```
Fetching gem metadata from http://rubygems.org/.........
Resolving dependencies...
Bundler could not find compatible versions for gem "nokogiri":
  In Gemfile:
    ruby-saml (0.9.1) ruby depends on
      nokogiri (~> 1.6.0) ruby

    nokogiri (1.5.11)
```

So it doesn't appear that Bundler is honoring the version-dependent dependency. One possible fix I can think of is to have just one dependency that is `>= 1.5.10`; however, I don't know whether you would consider that to be too broad.

Umofomia:
Fixes #198

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,12 @@ rvm:
   - 2.0.0
   - 2.1.5
   - ree
+gemfile:
+  - Gemfile
+  - gemfiles/nokogiri-1.5.gemfile
+matrix:
+  exclude:
+    - rvm: 1.8.7
+      gemfile: Gemfile
+    - rvm: ree
+      gemfile: Gemfile
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ gem 'ruby-saml', '~> 0.9.1'
 gem 'ruby-saml', :github => 'onelogin/ruby-saml'
 ```
 
-Using Bundler
+Using RubyGems
 
 ```sh
 gem install ruby-saml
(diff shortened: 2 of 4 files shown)


Lordnibbler:
This looks good @Umofomia, thank you for doing this.  

@luisvm @cgthornt can you please review?

cgthornt:
Looks sane to me :+1:

luisvm:
:+1: looking good

Lordnibbler:
thanks again @Umofomia

Umofomia:
Thanks for accepting this pull request, @Lordnibbler.

PR #201, opened 2015-03-13
Freeze and duplicate default security settings hash so that it doesn't get modified.

Umofomia:
I noticed that there was no option to generate signed metadata and decided to come up with an implementation for it (to come in a later pull request). While doing this, I looked at how signatures were currently being tested for other generated XML and noticed that the tests weren't actually asserting that the signature/digest methods were correct; e.g., they were doing:

```ruby
request_xml =~ /<ds:SignatureMethod Algorithm='http:\/\/www.w3.org\/2000\/09\/xmldsig#rsa-sha1'\/>/
```

instead of:

```ruby
assert_match %r[<ds:SignatureMethod Algorithm='http://www.w3.org/2000/09/xmldsig#rsa-sha1'/>], request_xml
```

Once I fixed these tests though, I noticed that the ones looking for a SHA1 signature were failing because the XML that was being generated used SHA256 that was set by a different test.

### Cause

The reason for this was because `Settings#initialize` simply sets `#security` to the `DEFAULTS[:security]` hash. When one of the tests set `security[:signature_method] = XMLSecurity::Document::SHA256`, it ended up changing the `DEFAULTS[:security]` hash itself, so it ends up affecting all `Settings` objects.

I added this test to `settings_test.rb` to clearly expose the issue:

```ruby
it "does not modify default security settings" do
  settings = OneLogin::RubySaml::Settings.new
  settings.security[:authn_requests_signed] = true
  settings.security[:embed_sign] = true
  settings.security[:digest_method] = XMLSecurity::Document::SHA256
  settings.security[:signature_method] = XMLSecurity::Document::SHA256

  new_settings = OneLogin::RubySaml::Settings.new
  assert_equal new_settings.security[:authn_requests_signed], false
  assert_equal new_settings.security[:embed_sign], false
  assert_equal new_settings.security[:digest_method], XMLSecurity::Document::SHA1
  assert_equal new_settings.security[:signature_method], XMLSecurity::Document::SHA1
end
```

This test will fail without the fix I describe in the following section.

### Fix

Typically when creating defaults, `#freeze` should be called on any values that are mutable. This will prevent inadvertent modification of default values. When initializing an object to use these default values, `#dup` should be called to duplicate any of the default values you want to be mutable without affecting the original default value.

### DIFF
--- a/lib/onelogin/ruby-saml/settings.rb
+++ b/lib/onelogin/ruby-saml/settings.rb
@@ -5,7 +5,10 @@ module OneLogin
         config = DEFAULTS.merge(overrides)
         config.each do |k,v|
           acc = "#{k.to_s}=".to_sym
-          self.send(acc, v) if self.respond_to? acc
+          if self.respond_to? acc
+            value = v.is_a?(Hash) ? v.dup : v
+            self.send(acc, value)
+          end
         end
         @attribute_consuming_service = AttributeService.new
       end
@@ -97,8 +100,8 @@ module OneLogin
       private
 
       DEFAULTS = {
-        :assertion_consumer_service_binding        => "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST",
-        :single_logout_service_binding             => "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect",
+        :assertion_consumer_service_binding        => "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST".freeze,
+        :single_logout_service_binding             => "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect".freeze,
         :compress_request                          => true,
         :compress_response                         => true,
         :security                                  => {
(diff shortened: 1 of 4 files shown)


pitbulk:
We will review that issue and metadata signature is in the queue. Right now we are focused on security and encryption support.

Umofomia:
@pitbulk I do already have a change ready that adds the metadata signature if you would consider it. However, it is blocked on having this bug fixed though.

Umofomia:
Rather, I should clarify that the metadata signature itself isn't blocked by this bug, but having the correct testing for it is what is blocked by this bug.

pitbulk:
@luisvm @Lordnibbler what do you think?

luisvm:
:+1: looks sane to me

Lordnibbler:
@Umofomia good catch, this looks good :+1: @cgthornt can you review for sanity?

phlipper:
:+1:

cgthornt:
:+1: looks sane

PR #204, opened 2015-03-17
Require ‘net/http’ library

rafaelgonzalez:
The only place where `Net::HTTP` is used is in `lib/ruby-saml/idp_metadata_parser.rb`, so this is where I require the `net/http` library.


Lordnibbler:
Thank you for this PR, @rafaelgonzalez

rafaelgonzalez:
Cheers :beers:

PR #206, opened 2015-03-17
use SecureRandom.uuid and encapsulate uuid creation

nddeluca:
I removed the uuid gem dependency in favor of using SecureRandom and added a util method encapsulating the implementation.  This reduces the gems footprint and uses a random uuid that is not based on the mac address and time of creation, which I think is important for use in authentication.

If the different uuid version is a problem (I'm not as familiar with the requirements of SAML), it may be better to use the `uuidtools` gem over the `uuid` gem as it's more up to date and doesn't pull in as many dependencies.

### DIFF
--- a/lib/onelogin/ruby-saml/utils.rb
+++ b/lib/onelogin/ruby-saml/utils.rb
@@ -1,6 +1,11 @@
 module OneLogin
   module RubySaml
     class Utils
+
+      def self.uuid
+        "_#{SecureRandom.uuid}"
+      end
+
       def self.format_cert(cert, heads=true)
         cert = cert.delete("\n").delete("\r").delete("\x0D")
         if cert
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -1,6 +1,5 @@
-require "uuid"
-
 require "onelogin/ruby-saml/logging"
+require "onelogin/ruby-saml/utils"
 
 module OneLogin
   module RubySaml
@@ -10,7 +9,7 @@ module OneLogin
       attr_reader :uuid # Can be obtained if neccessary
 
       def initialize
-        @uuid = "_" + UUID.new.generate
+        @uuid = OneLogin::RubySaml::Utils.uuid
       end
 
       def create(settings, params = {})
--- a/lib/onelogin/ruby-saml/idp_metadata_parser.rb
+++ b/lib/onelogin/ruby-saml/idp_metadata_parser.rb
@@ -1,5 +1,4 @@
 require "base64"
-require "uuid"
 require "zlib"
 require "cgi"
 require "net/http"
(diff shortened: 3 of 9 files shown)


lib/onelogin/ruby-saml/utils.rb
```diff
@@ -1,6 +1,11 @@
 module OneLogin
   module RubySaml
     class Utils
+
+      def self.uuid
+        "_#{SecureRandom.uuid}"
```
phlipper: This require should `require "securerandom"`.

phlipper:
I'm :+1: for this change, but `SecureRandom` is only available in Ruby 1.9+ and this gem has been holding on to support for 1.8, so until that changes there should be a fallback implementation.

lib/onelogin/ruby-saml/utils.rb
nddeluca: Ah right, thank you

nddeluca:
Sounds good, I guess I overlooked the ruby 1.8 support and will see if I can get a good fallback solution working.

TJM:
As I am a CentOS 6 user, with Ruby 1.8.7 as the de-facto standard Ruby, I can appreciate the need to look back to 1.8.7 still, but at some point (nearly 9 months after the official EOL), I think it may be time to "let it go"  <https://www.ruby-lang.org/en/news/2014/07/01/eol-for-1-8-7-and-1-9-2/> ?

I know that in my case, I am deploying to JRuby anyhow (which is why I am interested in getting rid of uuid), which I think is roughly equivalent to 1.9.3?

phlipper:
All Rubies `< 2.0.0` are EOL now.

TJM:
Sorry the "EOSL" on 1.9.3 just recently happened (like yesterday) <https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/> ...  I wasn't aware that it was here already. That should shake some things up :)

TJM:
Long story short, I think we can get by with SecureRandom.uuid and this gem: https://rubygems.org/gems/rubysl-securerandom to support 1.8.7 (right?)

TJM:
I hate to be a butt, but can we get this rebased, I would love to see this merged into the main repo, but its a bit old now. :)

Lordnibbler:
@nddeluca please? :smiley:  or worst case, @TJM you could rebase

subfusc:
This fix is needed for another bug that we are affected by, namely the fact that the way UUID is used today is not threadsafe. This can be shown using the code below:

```Ruby
(0..5).each do |j|
  Thread.new do
    (0..100000).each do |i|
      puts i if (i % 10000) == 0
      begin
        UUID.new.generate
      rescue Exception => e
        File.open(format('error_%i_%i.err', i, j), 'w') do |s|
          s.write(e)
        end
        raise e
      end
    end
  end
end
```

I rebased this and made it backwards compatible with 1.8.7 and fixed according to comments from @phlipper. Changes is in public branch on https://github.com/GyldendalDigital/ruby-saml/tree/use-secure-random-and-encapsulate-uuid-creation

How do you @nddeluca or upstream want me to proceed to get these changes into upstream?

Lordnibbler:
@subfusc sorry, I no longer maintain this gem :( you'll need to contact an engineer from Onelogin to get this issue addressed.

pitbulk:
Moved to #289

PR #207, opened 2015-03-18
Add ability to sign metadata.

Umofomia:
I also fixed the "embebed" typo that was copy-pasted across many of the comments.

### DIFF
--- a/lib/onelogin/ruby-saml/metadata.rb
+++ b/lib/onelogin/ruby-saml/metadata.rb
@@ -1,5 +1,3 @@
-require "rexml/document"
-require "rexml/xpath"
 require "uri"
 
 require "onelogin/ruby-saml/logging"
@@ -10,7 +8,6 @@ require "onelogin/ruby-saml/logging"
 # will be updated automatically
 module OneLogin
   module RubySaml
-    include REXML
     class Metadata
       def generate(settings)
         meta_doc = XMLSecurity::Document.new
--- a/lib/onelogin/ruby-saml/metadata.rb
+++ b/lib/onelogin/ruby-saml/metadata.rb
@@ -13,7 +13,7 @@ module OneLogin
     include REXML
     class Metadata
       def generate(settings)
-        meta_doc = REXML::Document.new
+        meta_doc = XMLSecurity::Document.new
         root = meta_doc.add_element "md:EntityDescriptor", {
             "xmlns:md" => "urn:oasis:names:tc:SAML:2.0:metadata"
         }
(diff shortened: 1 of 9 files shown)


test/metadata_test.rb
```diff
@@ -84,5 +84,42 @@ def setup
       assert_equal "Friendly Name", req_attr.attribute("FriendlyName").value
       assert_equal "Attribute Value", REXML::XPath.first(xml_doc, "//md:AttributeValue").text.strip
     end
+
```
Lordnibbler: it might be nice for cleanliness to instantiate the `settings` object in a `let` block, and do the setup for these tests in a `before` block

```ruby
describe "when the settings indicate to sign (embedded) the metadata" do
  let(:settings) { OneLogin::RubySaml::Settings.new }
  
  before do
    settings.issuer = "https://example.com"
    # ... etc
  end

  it "creates a signed metadata" do
    # your tests
  end
```

Umofomia: I was following the convention of the other tests in this file, but I can refactor this with the cleaner format. Should I refactor the existing tests as well?

Lordnibbler: @Umofomia If you've got the energy to do so, sure! That'd be wonderful! :)

Umofomia: @Lordnibbler Done. I only refactored the metadata tests since I noticed that @pitbulk still has #197 open and I didn't want to cause too many merge conflicts for him to resolve.

pitbulk:
@Umofomia,  Thanks, we are now focused on merging the #197 and adding encryption support, but will merge that soon ;)

lib/onelogin/ruby-saml/metadata.rb
```diff
@@ -85,6 +82,13 @@ def generate(settings)
         #  <md:XACMLAuthzDecisionQueryDescriptor WantAssertionsSigned="false" protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol"/>
 
         meta_doc << REXML::XMLDecl.new("1.0", "UTF-8")
+
+        # embed signature
+        if settings.security[:metadata_signed] && settings.private_key && settings.certificate && settings.security[:embed_sign]
```
pitbulk: The Signature on the Metadata element must be always embed (there is not other way) so we may do that independent of what value has the settings.security[:embed_sign].

Umofomia: Good point. I just pushed up a change that removes the check for `security[:embed_sign]`.

Umofomia:
Hmm... strange... TravisCI is failing, but the tests pass just fine on my machine. I'll try to see what's going on.

pitbulk:
@Umofomia the siganture_validation had a bug and I fixed it. No problem, I will fix your PR ;)

Umofomia:
@pitbulk Thanks, I just came to the same conclusion as well.

pitbulk:
@Umofomia, please review the PR #213

PR #210, opened 2015-03-20
Call assert in tests

bwillis:
I noticed there were a few test cases that didn't call assert, so I added them in.


Lordnibbler:
Thanks!

PR #211, opened 2015-03-24
Update base64 format check

phlipper:
* The main change here is adding `\\r` and `\\n` to the `gsub` in the check method. We were seeing a base64-encoded `SAMLResponse` parameter value come through with these extra escape characters and this solves is the issue.
* Extract the regexp for base64 format to a constant value to avoid additional unnecessary object allocations.
* Rename `is_base64?` to `base64_formatted?` since `rubocop` complains about methods starting with `is_`.

### DIFF
--- a/lib/onelogin/ruby-saml/saml_message.rb
+++ b/lib/onelogin/ruby-saml/saml_message.rb
@@ -14,6 +14,8 @@ module OneLogin
       ASSERTION = "urn:oasis:names:tc:SAML:2.0:assertion"
       PROTOCOL  = "urn:oasis:names:tc:SAML:2.0:protocol"
 
+      BASE64_FORMAT_REGEXP = %r{\A(([A-Za-z0-9+/]{4}))*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)\Z}
+
       def self.schema
         @schema ||= Mutex.new.synchronize do
           Dir.chdir(File.expand_path("../../../schemas", __FILE__)) do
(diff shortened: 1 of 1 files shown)


pitbulk:
:+1:
@Lordnibbler @luisvm what do you think?

lib/onelogin/ruby-saml/saml_message.rb
```diff
@@ -73,11 +75,11 @@ def encode(encoded)
...
       #
-      # The function is not strict and does allow newline. This is because some SAML implementations
-      # uses newline in the base64-encoded data, even if they shouldn't have (RFC4648).
-      def is_base64?(string)
-        string = string.gsub(/\r\n/, "").gsub(/\n/, "")
-        string.match(%r{\A(([A-Za-z0-9+/]{4}))*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)\Z})
+      # The function is not strict and allows newlines. This is because some
+      # SAML implementations use newlines in the base64-encoded data even if
+      # they shouldn't (RFC4648).
+      def base64_formatted?(string)
```
Lordnibbler: can you use YARD style comments here and explicitly document the return value as well?

Lordnibbler:
one small comment but looks good :P

Lordnibbler:
@pitbulk I had a comment here that was not addressed FWIW

pitbulk:
@Lordnibbler Yes, was already solved: 
https://github.com/onelogin/ruby-saml/commit/0552dc16acd8eba7dc14d361384beb261f4b5912

PR #205, opened 2015-03-17
Allow requirement of single files

rafaelgonzalez:
This is to reflect the [advertised behaviour in the README](https://github.com/onelogin/ruby-saml/tree/3264e2f97629fc8e459f64eb883c4b850abd8e92#getting-started).

You can manually test requirements of each file in `irb` (exiting `irb` between each) with:

- `require_relative 'lib/onelogin/ruby-saml/authrequest'`
- `require_relative 'lib/onelogin/ruby-saml/logoutrequest'`
- `require_relative 'lib/onelogin/ruby-saml/logoutresponse'`
- `require_relative 'lib/onelogin/ruby-saml/settings'`
- `require_relative 'lib/onelogin/ruby-saml/slo_logoutrequest'`
- `require_relative 'lib/onelogin/ruby-saml/slo_logoutresponse'`

PR #212, opened 2015-03-24
Rename library entry point

phlipper:
#205 introduced a regression when the entry point file was renamed: `lib/ruby-saml.rb → lib/onelogin/ruby-saml.rb`. This is causing problems with `bundler`, and also when integrating with other libraries such as `omniauth-saml`.


pitbulk:
:+1:

Lordnibbler:
@pitbulk this PR failed travis FYI, should ALWAYS wait until we're sure we're not merging a breaking change! 

thanks for this fix @phlipper

pitbulk:
@Lordnibbler, sorry, but It seems a Travis issue.

PR #214, opened 2015-03-25
Cleanup `SamlMessage` class

phlipper:
This includes some cleanups from working on other PRs. The commit messages should be self-explanatory.


lib/onelogin/ruby-saml/saml_message.rb
```diff
@@ -1,6 +1,6 @@
-require 'cgi'
-require 'zlib'
-require 'base64'
+require "cgi"
```
Lordnibbler: i'd opt for the opposite here, based on hound/rubocop -- double quote `"` only when using string interpolation, and single quote `'` everywhere else

Lordnibbler:
nice aside from my comment -- can you adjust the quotes? sorry to nitpick

phlipper:
@Lordnibbler no worries, I have updated the quotes.

Lordnibbler:
:+1:

PR #215, opened 2015-03-25
Fix Http-Redirect Signature: SignAlg and RelayState issue and algorithm calculator

pitbulk:
Now Signature algorithm 256, 384, 512 are supported on HTTP-Redirect Signatures (Tested with real IdPs). 

It was fixed a bug of mixing SHA1 and RSA_SHA1. algorithm and test were fixed.

We found a problem when creating the Signature, after some research we discovered that in some environments params['RelayState'] and params[:RelayState] are differents so Signatures were created without the RelayState parameter when keep sending a RelayState to the IdP, so the validation always failed. We added a compatibility line to fix that issue.

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -28,6 +28,8 @@ module OneLogin
 
       def create_params(settings, params={})
         params = {} if params.nil?
+        # Some ruby-saml versions uses :RelayState others use 'RelayState'
+        relay_state = params[:RelayState] || params['RelayState']
 
         request_doc = create_authentication_xml_doc(settings)
         request_doc.context[:attribute_quote] = :quote if settings.double_quote_xml_attribute_values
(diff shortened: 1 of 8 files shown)


lib/onelogin/ruby-saml/authrequest.rb
```diff
@@ -28,6 +28,8 @@ def create(settings, params = {})
 
       def create_params(settings, params={})
         params = {} if params.nil?
+        # Some ruby-saml versions uses :RelayState others use 'RelayState'
```
phlipper: What does this mean? Why can't the usage be consistent?

phlipper:
Does this PR reference existing GH issues? If so, which ones? If not, can you provide a little more context about this changeset? Thanks :)

lib/onelogin/ruby-saml/authrequest.rb
Lordnibbler: It might be that some people are using `HashWithIndifferentAccess` when they use this gem in Rails.

pitbulk:
@Lordnibbler Yes, that is the problem

@phlipper I updated the description of the PR and gonna add unit test to cover all Signature scenarios.

lib/onelogin/ruby-saml/authrequest.rb
phlipper: @Lordnibbler that is true when someone is using Rails, but this gem doesn't require or use Rails by default, so I would think the usage internally here should be consistent. Whichever method this gem chooses will then work fine since Rails will support either string or symbol keys. Am I missing something?

phlipper:
@pitbulk thanks for the additional context.

lib/onelogin/ruby-saml/authrequest.rb
Lordnibbler: True, perhaps we should be more explicit in the documentation of this method then.  Also, the assignment of `params = {} if params.nil?` seems unnecessary on line 30.

PR #222, opened 2015-04-17
no more silent failure fetching idp metadata

stevecrozz:
This is an attempt to avoid silent failure when fetching idp metadata. Currently the only way I am able to detect a failure is to reach into parser.document from outside and see if it's empty. With this PR, we get exceptions in two new cases:
1. When the URL has no scheme and we don't even attempt to fetch metadata, and
2. When the HTTP response is not an instance of Net::HTTPSuccess


lib/onelogin/ruby-saml/idp_metadata_parser.rb
```diff
@@ -62,7 +63,14 @@ def get_idp_metadata(url, validate_cert)
           get = Net::HTTP::Get.new(uri.request_uri)
           response = http.request(get)
           meta_text = response.body
+        elsif uri.scheme == nil
```
tony-spataro-rs: Shouldn't we validate what we assert? i.e. the elsif condition should become:

    elsif uri.scheme.nil? || uri.scheme !~ /https?/

stevecrozz: actually, it should probably just be an `else` so we catch all cases. i'll make that change

tony-spataro-rs:
:+1: :shipit:

pitbulk:
@bonyiii , @phene since you are using that feature on production, are you agree with this change?

Is it ok to raise an exception when the fetch fail?

I'm scared of those integration that are using this functionality and does not expect an Exception.

Maybe add a [soft parameter](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L189) and raise or not based on that. Default value will be not raise to keep compatibility issues. What do you think @stevecrozz and @tony-spataro-rs?

stevecrozz:
I think it should be fine to make changes as long as they're published in a new version and documented. But I would understand if you don't agree and I'm happy to make this new case a non-default.

phene:
My team uses another means for downloading the metadata and only provides `RubySaml` with the content when it's present, so this change should have no effect on us.

bonyiii:
I'm ok with this change.

tony-spataro-rs:
This seems like a valuable change in the long run; although I can understand the desire not to break compatibility, we should make things work correctly before the gem reaches v1.0. Two options:

-- If the user population are ok with the change, we should make it now (and add something to release notes). It's a change that only affects corner cases (metadata URL is invalid) that presumably would not result in successful SAML.

-- If someone has concerns, we could instrument the code so that the exception is raised if the gem version is >= 1.0

It sounds to me like option #1 is better; for my use case, an exception is better than silent failure.

Lordnibbler:
This change alters the behavior of this method, would be wise to at least add Yard documentation to the `parse_remote` method

stevecrozz:
@Lordnibbler I amended my commit, adding a new @raise to the yardoc for parse_remote and made an improvement there also. Since the YARDoc is identical for both parse_remote and get_idp_metadata, I was able to use "see" to reference one method's yardoc from the other method, and I verified that this generates the correct docs when running yard.

Lordnibbler:
@stevecrozz isn't `see` used like `@see`?

http://www.rubydoc.info/gems/yard/file/docs/Tags.md#see

stevecrozz:
@Lordnibbler That's what I thought at first, but @see just adds a link in the docs to the other method whereas @params (see ...) inlines the documatation as if you had written in here instead of "over there". I found it here and it does seem to work if we want the same docs in both places.
https://gist.github.com/chetan/1827484#methods

I can change it if you just want to "reference" the other docs though.

Lordnibbler:
ah thats nice! :+1:

PR #227, opened 2015-04-30
Save bindings specified in IdP metadata

justingaylor:
This change adds two new fields to the settings object:
- `idp_sso_target_url_bindings`
- `idp_slo_target_url_bindings`

These new fields are populated from the IdP metadata and are Arrays of Strings whose values are the Binding descriptors (e.g. `"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"`) in the order they were specified in the metadata file.

Signed-off-by: Justin Gaylor <justin@rightscale.com>


pitbulk:
Thanks for the contribution, right now the toolkit only support the HTTP-Redirect binding.

stevecrozz:
Can you clarify your comment @pitbulk? Are you meaning that this change is not welcome? or that you want to discuss it?

pitbulk:
@stevecrozz 

means that right now, we only support HTTP-POST for ACS endpoint and HTTP-Redirect for the rest.
So I believe that this commit is not necessary at this moment since there are no alternatives for the bindings and adding 2 extra parameters to the settings only turn it more complex. Are you disagree?

tony-spataro-rs:
That is a valid to omit the two extra attributes from settings; however, it's quite easy to support post and redirect for the ACS endpoint; the app we are building does so, using the gem as a basis. We have two reasons to care about the bindings in the settings objet:

- When we generate SP metadata (using the gem), we have no way to include information about the fact that we support both bindings. IdPs must guess at which bindings we support.

- When we consume IdP metadata, we would like to know which bindings are claimed in the IdP metadata so we can persist this to our DB. Currently we re-parse the metadata and grab the information ourselves, but that's a bit dirty.

tony-spataro-rs:
Something else to consider, is that it's very easy to adapt the gem to support HTTP-POST binding for AuthnRequest. We did it with ten lines of code:

        uri = request.create(settings) # calling ruby-saml
        query_string = CGI.parse(URI.parse(uri).query)
        form_data = {
          method: 'post',
          sso_target_url: settings.idp_sso_target_url,
          relay_state: relay_state,
          saml_request: query_string['SAMLRequest'],
          sig_alg: query_string['SigAlg'],
          signature: query_string['Signature']
        }
        render 'saml/interstitial', form_data

We use a similar technique to support HTTP-Redirect binding for our ACS URL.

pitbulk:
>When we consume IdP metadata, we would like to know which SSO-endpoint bindings are claimed in the IdP metadata so we can persist this to our DB. Currently we re-parse the metadata and grab the information ourselves, but that's a bit dirty.

I integrated [single_signon_service_binding](https://github.com/onelogin/ruby-saml/commit/2e60902adf7748cdb323577051dd915cac2f1a30#diff-224905527907deba9ffd52ff7ab83308R124) and [single_logout_service_binding](https://github.com/onelogin/ruby-saml/commit/2e60902adf7748cdb323577051dd915cac2f1a30#diff-224905527907deba9ffd52ff7ab83308R154) at idp_metadata_parser.rb at https://github.com/onelogin/ruby-saml/commit/2e60902adf7748cdb323577051dd915cac2f1a30

so you will be able to consume IdP metadata

>When we generate SP metadata (using the gem), we have no way to include information about the fact that we support both bindings. IdPs must guess at which bindings we support.

[Right now we offer the binding information on the metadata](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/metadata.rb#L69). [SP binding is set at the settings](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/settings.rb#L147) but you may now right now is supported HTTP-POST for ACS and HTTP-Redirect for SLO.

We are trying to support the [saml2int](http://saml2int.org/profile/current/) profile:
* The <saml2p:AuthnRequest> message issued by a Service Provider MUST be communicated to the Identity Provider using the HTTP-REDIRECT binding
* The <saml2p:Response> message issued by an Identity Provider MUST be communicated to the Service Provider using the HTTP-POST binding

PR #231, opened 2015-05-05
Refactor HTTP-Redirect Sign method, Move test data to right folder

pitbulk:
Splitted from #197


Lordnibbler:
More duplication here in `authrequest.rb`, `logoutrequest.rb`, and `slo_logoutresponse.rb`.  Would really love to see this refactored into a parent class

PR #232, opened 2015-05-05
Improve validations: Store the causes in the errors array, code refactor

pitbulk:
Improve validations:
* Store the causes in the errors array
* Clean code

Improve response_test, test_helper and xml_security:
* Avoid random names (vars and files)
* Clean code

Some code splitted from #197


Lordnibbler:
There is a tremendous amount of duplication in `logoutresponse.rb`, `response.rb`, and `slo_logoutrequest.rb`.  Would it make sense to make a superclass and have these inherit from it?

test/response_test.rb
```diff
@@ -68,10 +68,17 @@ class RubySamlTest < Minitest::Test
     end
 
     describe "#validate_structure" do
-      it "raise when encountering a condition that prevents the document from being valid" do
+      it "false when encountering a mailformed element that prevents the document from being valid" do
+        response = OneLogin::RubySaml::Response.new(response_document_2)
```
Lordnibbler: move this to a `let()` block

test/response_test.rb
```diff
@@ -68,10 +68,17 @@ class RubySamlTest < Minitest::Test
...
+        response.send(:validate_structure, true)
+        assert response.errors.include? "Invalid SAML Response. Not match the saml-schema-protocol-2.0.xsd"
+      end
+
+      it "raise when encountering a mailformed element that prevents the document from being valid" do
         response = OneLogin::RubySaml::Response.new(response_document_2)
```
Lordnibbler: move this to a `let()` block

pitbulk:
@Lordnibbler I cleaned the response_test and test_helper. Is ok if I merge it?

test/response_test.rb
```diff
@@ -363,29 +328,24 @@ class RubySamlTest < Minitest::Test
 
     describe "#session_expires_at" do
       it "extract the value of the SessionNotOnOrAfter attribute" do
-        response = OneLogin::RubySaml::Response.new(response_document)
         assert response.session_expires_at.is_a?(Time)
 
-        response = OneLogin::RubySaml::Response.new(response_document_2)
-        assert_nil response.session_expires_at
+        assert_nil response_without_attributes.session_expires_at
```
luisvm: imo, this should go in a different test case, stating something like: `it "returns nil when the response does not have attributes"`

test/response_test.rb
```diff
@@ -393,8 +353,8 @@ class RubySamlTest < Minitest::Test
     describe '#xpath_first_from_signed_assertion' do
       it 'not allow arbitrary code execution' do
         malicious_response_document = fixture('response_eval', false)
-        response = OneLogin::RubySaml::Response.new(malicious_response_document)
-        response.send(:xpath_first_from_signed_assertion)
+        malicious_response = OneLogin::RubySaml::Response.new(malicious_response_document)
+        malicious_response.send(:xpath_first_from_signed_assertion)
         assert_equal($evalled, nil)
```
luisvm: is it possible we can redo this spec case using a different approach? ( I personally don't like using global vars )

pitbulk: @luisvm what is your proposal?

PR #234, opened 2015-05-07
Add explicit support for JRuby

ylansegal:
Dear maintainers, I would like to explicitly add support for JRuby to this gem. I was able to make the necessary changes, as follows:

1. Some stubs where causing some issues that are resolved by either explicitly un-stubbing or stubbing a particular instance, as opposed to any_instance
2. Nokogiri has it's own java extension different from the one in MRI. The 1.5 series has some bugs that have
been resolved in 1.6 series, so I use that version when in JRuby
3. Nokogiri has different behavior when validating XML schemas in MRI and JRuby. I already raised the issue there (https://github.com/sparklemotion/nokogiri/issues/1282). To get around it, I removed the memoization that was earlier added as a performance improvement. Is this acceptable?
4. The `Nokogiri::XML::Node#canonicalize` method in Nokogiri also behaves differently in JRuby vs MRI. However, that is easily solved by passing in `nil` (which is the [default parameter](http://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Node#canonicalize-instance_method)) instead of `[]` for the inclusive_namespaces.

I believe 3 is the only real change in behavior for the gem. I'd appreciate your feedback.

Travis build: https://travis-ci.org/ylansegal/ruby-saml/builds/61692511

Thanks


pitbulk:
I'm glad with with the fact of supporting JRuby, and I'm ok with the code. @luisvm, @Lordnibbler what do you think?

ylansegal:
I have updated this PR, because of conflicts with the recent changes in master. What are the chances of this being merged?

Also, besides the Nokogiri mentioned above. JRuby 1.7.20 also has some change in behavior (in `jruby-openssl`) that breaks the tests. I have created an issue for them: https://github.com/jruby/jruby-openssl/issues/42

Thanks,

ylansegal:
JRuby 9000 is out, and jruby-openssl 0.9.8 fixes some X509 cert parsing issues. I updated the pull request. I am aware that there are merge conflicts: I will gladly resolve them again if there is interest from the committers in merging this PR.

pitbulk:
@ylansegal

Apologies for not merging your PR on time, I think that supporting JRuby makes sense since it not requires much changes and can help some of the java-customers. 

Please fix conflicts, we will review the changes, test and include it on the master branch.

ylansegal:
@pitbulk I've fixed the merge conflicts and added support for JRuby-1.7.21 and JRuby-9.0.0.0. 

Thank you,

pitbulk:
We will review it and get it merged this week. Thanks for contributing!.

keaplogik:
@ylansegal does the issue you have with schema validate relate to not being able to validate on hard validation?

I do get an invalid_ticket error in JRuby but not on MRI. 

```ruby
response = RubySaml::Response.new(request.params['SAMLResponse'], options)

unless response.is_valid?
    raise ValidationError.new(response.errors.join('\n'))
end
```

ylansegal:
@keaplogik Are you seeing issues after this PR was merged? All the ones I was having where solved.

The issue with the schema validation that I mentioned in the original post was solved by not memoizing the `Nokogiri::XML::Schema` object, which triggers a difference in behavior between Nokogiri in MRI vs JRuby. 

I am not sure what is causing your `invalid_ticket`

keaplogik:
OK thanks. Looks to be a separate issue I may need to do some more verification on my test. Looks like a digest mismatch error.

Issue #220, opened 2015-04-01
document.validate fails if ds:x509certificate is not embedded in the SAML Response


TJM:
We are building an app that is going to do SSO with Ping Federate using ruby-saml, and I am told that the IdP is emitting valid SAMLv2 responses. However, the responses do not contain  the ds:x509certificate stanza. The certificate *is* available in the metadata.xml file, and it matches the signature in the SAML response. I am not sure why the cert would have to be embedded in the response, when it is already available in the metadata. We do have an option for "anchored" response that does include the certificate. So this is partially to help someone else in the future who runs into this same issue and partially to figure out why the certificate would be required in the SAML response. Our initial problem is solved, but it has left me wondering why. Hopefully someone can help.

~tommy

TJM:
Quote from: http://stackoverflow.com/questions/1703301/saml-why-is-the-certificate-within-the-signature

> SAML responses do not require including the public key for that signature. Section 5.4.5 of the SAML2  spec states "XML Signature defines usage of the <ds:KeyInfo> element. SAML does not require the use of <ds:KeyInfo>, nor does it impose any restrictions on its use. Therefore, <ds:KeyInfo> MAY be absent." You can verify the signature if the public key has been provided to you through other means, e.g. stored in your local certificate store prior to implementing the SAML consumer. –  Sam.Rueby Nov 5 '14 at 16:01

pitbulk:
The ruby-saml gem expects a Signed SAML Response, This Signature may carry a cert.
In our signature validation checking we extract the x509certificate from the SAMLResponse and compare it with the settings.idp_cert or the settings.idp_fingerprint.

Most of the Service Providers implementations works in this way.

The main reason is that most SAML implementations delegate on XMLSignature implementations that requires the KeyInfo element, but I'm agree that the spec says that SAML does not require it (. Section 5.4.5). 

Other reason for doing that is for avoiding DOS attacks (saml-sec-consider 5.2.1 Denial of Service).
Is fast/easy discard a SAMLResponse without cert or an invalid cert than verify the signature.

TimMoore:
@pitbulk would you consider accepting a pull request that makes it optional in the response document if `settings.idp_cert` is defined?

pitbulk:
@TimMoore, I posted why we implemented all our toolkits requiring the x509certificate data.

I think is easier configuring the Identity Provider side to inject the ds:x509certificate.
99% of Identity Provider implementations do that.

TJM:
[...]

PR #236, opened 2015-05-13
Force assertions to be signed.

patcheng:
address https://github.com/onelogin/ruby-saml/issues/220

When processing the response, the toolkit rejects anything that's unsigned. So the Metadata XML should always be WantAssertionsSigned="true"

But in https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/metadata.rb#L29

WantAssertionsSigned is set to true only when settings.idp_cert_fingerprint or settings.idp_cert is set. In our workflow at least, user will most likely generates the metadata before setting the idp certs.

### DIFF
--- a/lib/onelogin/ruby-saml/metadata.rb
+++ b/lib/onelogin/ruby-saml/metadata.rb
@@ -25,8 +25,8 @@ module OneLogin
         sp_sso = root.add_element "md:SPSSODescriptor", {
             "protocolSupportEnumeration" => "urn:oasis:names:tc:SAML:2.0:protocol",
             "AuthnRequestsSigned" => settings.security[:authn_requests_signed],
-            # However we would like assertions signed if idp_cert_fingerprint or idp_cert is set
-            "WantAssertionsSigned" => !!(settings.idp_cert_fingerprint || settings.idp_cert)
+            # However we always want assertions signed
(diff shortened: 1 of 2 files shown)


pitbulk:
Thanks for comment that, I will provide a solution on that since we don't want to force always the WantAssertionsSigned="true". Maybe I will add new parameters to the 'security' settings.

patcheng:
Here is how I interpret the specs:

according to the SAML-core 2.0 section 5.3, Assertion elements is signed, or when the Assertion element is contained in an element (Response) that's signed.

according saml-metadata-2.0, WantAssertionsSigned means the service provider wants the Assertion element to be signed. But they didn't specified if the Signature element has to be inside of the Assertion element, or through "Signature Inheritance". 

I supposed you could interpret WantAssertionsSigned to Assertion has to have its own Signature.

But if an IdP signs the Response instead of the Assertion, the Assertion is technically signed anyway.

Just want to make sure I am not missing something

pitbulk:
@patcheng  review what I already implemented in the php-saml toolkit.
https://github.com/onelogin/php-saml/blob/master/advanced_settings_example.php

If you see I have settings to configure exactly what the SP expects.

patcheng:
ok.

PR #237, opened 2015-05-22
ugly print metadata by default

stevecrozz:
As the method comment describes, SP metadata signatures cannot be
verified if they are pretty printed. For this reason, I contend that
pretty-print as a default is not sane. Certain IdPs will refuse to parse
signed metadata unless the signature can be verified (notably AD FS).

Changing the default will have no effect for users who explicitly set the value, and it should not have any adverse effect beyond impairing human readability of the raw xml and should only improve interoperability for those users who sign their metadata documents.


pitbulk:
Agree.

PR #241, opened 2015-06-11
Add decrypt support.

pitbulk:
* Be able to decrypt EncryptedAssertions


pitbulk:
To end the PR, I need to add support fo EncryptedID (Encrypted NameID).

meanwhile, @luisvm and @Lordnibbler can you review those commits?

lib/onelogin/ruby-saml/response.rb
```diff
@@ -51,6 +53,21 @@ def initialize(response, options = {})
 
         @response = decode_raw_saml(response)
         @document = XMLSecurity::SignedDocument.new(@response, @errors)
+
+        if assertion_encrypted?
```
luisvm:
```ruby
return unless assertion_encrypted?
...
```
that will save you from indenting the following lines

lib/onelogin/ruby-saml/response.rb
```diff
@@ -51,6 +53,21 @@ def initialize(response, options = {})
...
+          end
+
+          # Marshal at Ruby 1.8.7 throw an Exception
+          if RUBY_VERSION < "1.9"
+            @decrypted_document = XMLSecurity::SignedDocument.new(@response, @errors)
+          else
+            @decrypted_document = Marshal.load(Marshal.dump(@document))
+          end
+
+          @decrypted_document = document_with_decrypted_assertion
```
luisvm: why do we need to store `@decrypted_document` in the `if` above and then write to the same variable here again?

pitbulk: @luisvm  @decrypted_document is a copy of the original saml response that will be modified at the document_with_decrypted_assertion method, that is why I need to create a copy of this object and save it.

lib/onelogin/ruby-saml/response.rb
```diff
@@ -51,6 +53,21 @@ def initialize(response, options = {})
...
         @document = XMLSecurity::SignedDocument.new(@response, @errors)
+
+        if assertion_encrypted?
+          if @settings.nil? || !@settings.get_sp_key
+            validation_error('An EncryptedAssertion found and no SP private key found on the settings to decrypt it. Be sure you provided the :settings parameter at the initialize method')
+          end
+
+          # Marshal at Ruby 1.8.7 throw an Exception
+          if RUBY_VERSION < "1.9"
+            @decrypted_document = XMLSecurity::SignedDocument.new(@response, @errors)
```
luisvm: in this case, this would just be
```ruby
@decrypted_document = @document
```

since that was parsed in line 55

lib/onelogin/ruby-saml/response.rb
luisvm: is there a reason to store both the encrypted and decrypted documents?
I vote to just decrypt it, parse it and assign it to `@document`

in the remote case we need the encrypted version, we have it inside `@response`

pitbulk: We need both version.

When we check the signature validation of the saml response for example, we need the original response, and later we need the decrypted to get the elements.

luisvm:
:+1: looks good.
could you please get more reviewers before merging?

test/metadata_test.rb
```diff
@@ -70,25 +70,42 @@ class MetadataTest < Minitest::Test
...
 
-      it "generates Service Provider Metadata with X509Certificate" do
+      it "generates Service Provider Metadata with AuthnRequestsSigned" do
+        settings.security[:authn_requests_signed] = true
         assert_equal "true", spsso_descriptor.attribute("AuthnRequestsSigned").value
         assert_equal ruby_saml_cert.to_der, cert.to_der
       end
+
+      it "generates Service Provider Metadata with X509Certificate for sign and encrypt" do
+        assert_equal 2, key_descriptors.length
```
daniel-g: if there were 3... would that change anything in this spec? otherwise, we don't need the line

test/response_test.rb
```diff
@@ -35,7 +35,7 @@ class RubySamlTest < Minitest::Test
     let(:response_invalid_subjectconfirmation_nb) { OneLogin::RubySaml::Response.new(read_invalid_response("invalid_subjectconfirmation_nb.xml.base64")) }
     let(:response_invalid_subjectconfirmation_noa) { OneLogin::RubySaml::Response.new(read_invalid_response("invalid_subjectconfirmation_noa.xml.base64")) }
     let(:response_invalid_signature_position) { OneLogin::RubySaml::Response.new(read_invalid_response("invalid_signature_position.xml.base64")) }
-
+    let(:response_encrypted_nameid) { OneLogin::RubySaml::Response.new(response_document_encrypted_nameid) }
```
daniel-g: nameid to name_id. One underscore changes averything

test/metadata_test.rb
pitbulk: We testing that there are 2 and only 2.

test/response_test.rb
```diff
@@ -852,9 +852,237 @@ class RubySamlTest < Minitest::Test
...
+          assert signed_response.is_valid?
+        end
+        assert_empty signed_response.errors
+      end
+    end
+
+    describe "retrieve nameID" do
+      it 'able when nameID inside the assertion' do
+        response_valid_signed.settings = settings
+        assert_equal "test@onelogin.com", response_valid_signed.name_id
```
daniel-g: so.. name_id or nameid ?

test/response_test.rb
```diff
@@ -852,9 +852,237 @@ class RubySamlTest < Minitest::Test
...
+          response_encrypted_nameid.settings = settings
+          assert_raises(OneLogin::RubySaml::ValidationError, "An EncryptedID found and no SP private key found on the settings to decrypt it") do
+            assert_equal "test@onelogin.com", response_encrypted_nameid.name_id    
+          end
+      end
+
+      it 'able when encryptID inside the assertion and settings has the private key' do
+        settings.private_key = ruby_saml_key_text
+        response_encrypted_nameid.settings = settings
+        assert_equal "test@onelogin.com", response_encrypted_nameid.name_id
```
daniel-g: this one really confuses me... so nameid has attributes?

test/response_test.rb
```diff
@@ -852,9 +852,237 @@ class RubySamlTest < Minitest::Test
...
+    end
+
+  end
+
+  describe '#is_assertion_encrypted?' do
+    it 'return false if none encrypted assertion is found' do
+      settings.certificate = ruby_saml_cert_text
+      settings.private_key = ruby_saml_key_text
+      response = OneLogin::RubySaml::Response.new(response_document_valid_signed, :settings => settings)
+      assert !response.send(:assertion_encrypted?)
```
daniel-g: is this a private method?

test/response_test.rb
```diff
@@ -852,9 +852,237 @@ class RubySamlTest < Minitest::Test
...
+
+    before do
+      settings.idp_cert_fingerprint = 'EE:17:4E:FB:A8:81:71:12:0D:2A:78:43:BC:E7:0C:07:58:79:F4:F4'
+      settings.issuer = 'http://rubysaml.com:3000/saml/metadata'
+      settings.assertion_consumer_service_url = 'http://rubysaml.com:3000/saml/acs'
+      settings.certificate = ruby_saml_cert_text
+      settings.private_key = ruby_saml_key_text
+    end
+
+    it 'able when signed_message_encrypted_unsigned_assertion' do
```
daniel-g: it ables what?

daniel-g: if blank... then enables ?

luisvm: yeah, you probably meant `it 'is enabled when...'`

luisvm: and same with the examples below

test/response_test.rb
```diff
@@ -852,9 +852,237 @@ class RubySamlTest < Minitest::Test
...
+        response = OneLogin::RubySaml::Response.new(signed_message_encrypted_unsigned_assertion, :settings => settings)        
+
+        encrypted_assertion_node = REXML::XPath.first(
+          response.document,
+          "(/p:Response/EncryptedAssertion/)|(/p:Response/a:EncryptedAssertion/)",
+          { "p" => "urn:oasis:names:tc:SAML:2.0:protocol", "a" => "urn:oasis:names:tc:SAML:2.0:assertion" }
+        )
+        response.settings.private_key = nil
+
+        error_msg = "An EncryptedAssertion found and no SP private key found on the settings to decrypt it"
```
daniel-g: the message can change, I think you should only test the exception

test/response_test.rb
```diff
@@ -852,9 +852,237 @@ class RubySamlTest < Minitest::Test
...
+    describe "retrieve nameID" do
+      it 'able when nameID inside the assertion' do
+        response_valid_signed.settings = settings
+        assert_equal "test@onelogin.com", response_valid_signed.name_id
+      end
+
+      it 'unable when encryptID inside the assertion but no private key' do
+          response_encrypted_nameid.settings = settings
+          assert_raises(OneLogin::RubySaml::ValidationError, "An EncryptedID found and no SP private key found on the settings to decrypt it") do
+            assert_equal "test@onelogin.com", response_encrypted_nameid.name_id
```
daniel-g: @pitbulk

test/response_test.rb
pitbulk: you need to read the describe...

retrieve nameID and attributes from encrypted assertion ... able when signed_message_encrypted_unsigned_assertion

test/response_test.rb
pitbulk: It should not change, if change for a reason, we may update the tests related.
Provide standard errors is something that help developers ;)

lib/onelogin/ruby-saml/response.rb
```diff
@@ -51,6 +53,21 @@ def initialize(response, options = {})
 
         @response = decode_raw_saml(response)
         @document = XMLSecurity::SignedDocument.new(@response, @errors)
+
+        return unless assertion_encrypted?
```
daniel-g: If the code after this line is for encrypt the assertion, you should create a method for that.

test/response_test.rb
daniel-g: I don´t mean it is not important, I mean the exception is enough for the test, plus you won't have to change the test if the description changes

test/response_test.rb
pitbulk: Yes, is a private method, only used internally, but wanted to add an specific test for it.

daniel-g: better add specs for the public methods that uses it

pitbulk: I also did it

daniel-g: then you don't need theses specs. The others will let you know if this method fails. The problem writing specs for private methods is that these kind of methods were made to be changed internally, so it doesn't make sense to write them.

test/response_test.rb
pitbulk: lol, is not nameid.name_id !

I have a response_encrypted_nameid that is a response that has an encrypted nameid (I gave this name to difference this response from the rest, and I wanted to get the nameid attribute of this response.   What you suggest?

daniel-g: no suggestion at all I think hehe, you are following SAML, but I wonder if the name_id is actually a nameid

luisvm: he's probably referring to the fact that we have `nameid.nameid` or `nameid.name_id` doesn't matter if it has an underscore or not, it just seems weird to call `nameid` attribute on `nameid` element, you could probably use `nameid_element.nameid` or `subject_element.nameid` just to denote who's the parent

lib/onelogin/ruby-saml/response.rb
luisvm: **decrypt

pitbulk: I created a method for that: decrypt_assertion_from_document    but the problem is that I need to clone the document in order to later decrypt it (if I no clone, then when decrypting I change the original object). And due Marshal bugs on  Ruby 1.8.7 I need to offer an alternative re-parsing the response.  And I want to pass to the method the cloned object. Does it make sense?

test/response_test.rb
daniel-g: tongue-twister ... just follow a name convention... either nameid or name_id

lib/onelogin/ruby-saml/response.rb
luisvm: why not clone the object inside `decrypt_assertion_from_document` and pass `document` or `response` to that same method?

pitbulk: @luisvm, I think that your proposal turn the code more complex..and this is maybe what we are trying to avoid..

luisvm: @pitbulk well, initialize is already pretty complex

test/response_test.rb
pitbulk: Right now here, I think we only have response.name_id     but I named responses that has a Encrypted NameID element:  response_encrypted_nameid

name_id is the response method to retrieve the NameID attribute from a SAML Response.
Remember that this method is here since long time ago:
https://github.com/onelogin/ruby-saml/blob/0.7.3/lib/onelogin/ruby-saml/response.rb#L37

I think that the right thing is to name it nameid, that is how I named it at the [php](https://github.com/onelogin/php-saml/blob/4c2bba447b9271b605d05c6c54eeabd7f7e64500/lib/Saml/Response.php#L31) and [python](https://github.com/onelogin/python-saml/blob/master/src/onelogin/saml2/response.py#L299).

The problem is that we can't change the name of this method since is a method that is already used in 3rd party apps.   Maybe create both methods for compatibility issues and start to use in our code the response.nameid  ??

lib/onelogin/ruby-saml/response.rb
pitbulk: @luisvm could be that a solution?
https://gist.github.com/pitbulk/c143ddc83aeaefca3834

lib/onelogin/ruby-saml/response.rb
```diff
@@ -76,7 +93,12 @@ def is_valid?
       #
       def name_id
         @name_id ||= begin
-          node = xpath_first_from_signed_assertion('/a:Subject/a:NameID')
+          enc_node = xpath_first_from_signed_assertion('/a:Subject/a:EncryptedID')
```
daniel-g: I would complete the name to make it more readable. encrypted_node

test/response_test.rb
daniel-g: yes, an alieas could work

lib/onelogin/ruby-saml/response.rb
```diff
@@ -565,17 +594,18 @@ def validate_signature
       # @return [REXML::Element | nil] If any matches, return the Element
       #
       def xpath_first_from_signed_assertion(subelt=nil)
+        doc = decrypted_document.nil? ? document : decrypted_document
```
daniel-g: This line is repeated, maybe it worth a method?

lib/onelogin/ruby-saml/response.rb
```diff
@@ -586,20 +616,93 @@ def xpath_first_from_signed_assertion(subelt=nil)
...
+      # Checks if the SAML Response contains or not an EncryptedAssertion element
+      # @return [Boolean] True if the SAML Response contains an EncryptedAssertion element
+      #
+      def assertion_encrypted?
+        encrypted_node = REXML::XPath.first(
+          document,
+          "(/p:Response/EncryptedAssertion/)|(/p:Response/a:EncryptedAssertion/)",
+          { "p" => PROTOCOL, "a" => ASSERTION }
+        )
+        !encrypted_node.nil?
```
daniel-g: I would make this simpler:
```
! REXML::XPath.first(
          document,
          "(/p:Response/EncryptedAssertion/)|(/p:Response/a:EncryptedAssertion/)",
          { "p" => PROTOCOL, "a" => ASSERTION }
        ).nil?
```

lib/onelogin/ruby-saml/response.rb
```diff
@@ -586,20 +616,93 @@ def xpath_first_from_signed_assertion(subelt=nil)
...
+      end
+
+      # Decrypts an EncryptedAssertion element
+      # @param encrypted_assertion_node [REXML::Element] The EncryptedAssertion element
+      # @return [REXML::Document] The decrypted EncryptedAssertion element
+      #
+      def decrypt_assertion(encrypted_assertion_node)
+        if settings.nil? || !settings.get_sp_key
```
daniel-g: this can be expressed with an eager return:
```
if settings.nil? || !settings.get_sp_key
  return validation_error('An EncryptedAssertion found and no SP private key found on the settings to decrypt it')
end
<Then all the code>
```

lib/onelogin/ruby-saml/response.rb
```diff
@@ -586,20 +616,93 @@ def xpath_first_from_signed_assertion(subelt=nil)
...
+          doc.root[0]
+        end
+      end
+
+      # Decrypts an EncryptedID element
+      # @param encryptedid_node [REXML::Element] The EncryptedID element
+      # @return [REXML::Document] The decrypted EncrypedtID element
+      #
+      def decrypt_nameid(encryptedid_node)
+        if settings.nil? || !settings.get_sp_key
```
daniel-g: eager return here too

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -86,6 +90,63 @@ def self.status_error_msg(error_msg, status_code = nil, status_message = nil)
 
         error_msg
       end
+
+      # Obtains the decrypted string from an Encrypted element
```
daniel-g: from an Encrypted node element in XML

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -86,6 +90,63 @@ def self.status_error_msg(error_msg, status_code = nil, status_message = nil)
...
+      # @param private_key    [OpenSSL::PKey::RSA] The Service provider private key
+      # @return [String] The decrypted data
+      def self.decrypt_data(encrypted_node, private_key)
+        encrypt_data = REXML::XPath.first(encrypted_node, "./xenc:EncryptedData", { 'xenc' => XENC })
+        symmetric_key = retrieve_symmetric_key(encrypt_data, private_key)
+        cipher_value = REXML::XPath.first(encrypt_data, "//xenc:EncryptedData/xenc:CipherData/xenc:CipherValue", { 'xenc' => XENC })
+        node = Base64.decode64(cipher_value.text)
+        enc_method = REXML::XPath.first(encrypt_data, "//xenc:EncryptedData/xenc:EncryptionMethod", { 'xenc' => XENC })
+        algorithm = enc_method.attributes['Algorithm']
+        assertion_plaintext = retrieve_plaintext(node, symmetric_key, algorithm)
```
daniel-g: do you need assertion_plaintext to have value at the end of the method?

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -86,6 +90,63 @@ def self.status_error_msg(error_msg, status_code = nil, status_message = nil)
...
+        algorithm = enc_method.attributes['Algorithm']
+        assertion_plaintext = retrieve_plaintext(node, symmetric_key, algorithm)        
+      end
+
+      # Obtains the symmetric key from the EncryptedData element
+      # @param encrypt_data [REXML::Element]     The EncryptedData element
+      # @param private_key [OpenSSL::PKey::RSA] The Service provider private key
+      # @return [String] The symmetric key
+      def self.retrieve_symmetric_key(encrypt_data, private_key)
+        encrypted_symmetric_key_element = REXML::XPath.first(encrypt_data, "//xenc:EncryptedData/ds:KeyInfo/xenc:EncryptedKey/xenc:CipherData/xenc:CipherValue", { "ds" => DSIG, "xenc" => XENC })
```
daniel-g: can you split this method call into many lines?

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -86,6 +90,63 @@ def self.status_error_msg(error_msg, status_code = nil, status_message = nil)
...
+        retrieve_plaintext(cipher_text, private_key, algorithm)        
+      end
+
+      # Obtains the deciphered text
+      # @param cipher_text [String]   The ciphered text
+      # @param symmetric_key [String] The symetric key used to encrypt the text
+      # @param algorithm [String]     The encrypted algorithm
+      # @return [String] The deciphered text
+      def self.retrieve_plaintext(cipher_text, symmetric_key, algorithm)
+        case algorithm
```
daniel-g: This method seems to belong to its own class. If you feel like it won't grow in time, then keep it here.

daniel-g:
@pitbulk please look at the conversation, some of them doesn't make sense, so please tell me if I'm thinking it wrong :)

test/response_test.rb
luisvm: `is possible when signed_message....` that's what `it`s are for, they should read fine with `it` and the following sentence
probably `is possible` are the words you were looking for =P

test/response_test.rb
luisvm: oh nono, `nameid` should be fine, one concern was to be consistent on whether we should be using `nameid` or `name_id`.
My concern was that the variable `response_encrypted_nameid` contains an instance of `OneLogin::RubySaml::Response` so I would rename `response_encrypted_nameid` to `encrypted_response`, but this is a nitpick. up to you

pitbulk: Luis, what is encrypted on that response is the NameID, not the assertion.

luisvm: @pitbulk this has become a big deal out of a minor issue, please ignore

lib/onelogin/ruby-saml/response.rb
pitbulk: I will refactor both methods ;)

daniel-g:
:+1:

PR #243, opened 2015-06-26
Fix metadata validation errors

soupmatt:
There are xml schema validation errors with the service provider metadata xml that is generated by `OneLogin::RubySaml::Metadata#generate`. The first is that `SingleLogoutService` isn't allowed to to have `isDefault` and `index` attributes. The other is that the `AttributeValue` element comes from the assertion schema, not the metadata schema. See the diff for the details.

I also added code that validates the xml produced by each test against the schema.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ coverage
 rdoc
 pkg
 Gemfile.lock
+gemfiles/*.lock
 .idea/*
 lib/Lib.iml
 test/Test.iml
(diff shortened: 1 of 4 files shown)


pitbulk:
You are right, [SingleLogoutService](https://github.com/onelogin/ruby-saml/blob/master/lib/schemas/saml-schema-metadata-2.0.xsd#L219) is EndpointType not IndexedEndpointType, so no isDefault / index allowed.

Thanks for contributing that

Lordnibbler:
seems sane :+1:

PR #251, opened 2015-07-08
Support qualified and unqualified NameID in attributes

balmas:
Our use case has the IdP returning the eduPersonTargetedId in the attributes to use as an alternative to the NameID in the subject, which in our case is transient and can't be used as stable identifier for the user. I believe this is a valid SAML2 scenario (we are using it with Shiboleth).  

The response.attributes method was ignoring complex AttributeValues.  This change allows it to take them into account only for this particular case of NameID, and also adds support for the NameQualifier on the NameID value, returning that along with the value, separated by "/" to create a fully qualified URI as the value.

Tests are included.  Would be great to get this change included in the master ruby-saml code base.

Thanks!


balmas:
Thanks!!

PR #253, opened 2015-07-10
Handle valid but uncommon dsig block with no URI in the reference

knightlabs:
Encountered a situation in which the ds:Reference block in a message signature was empty; published as kl-ruby-saml-0.0.2 if anyone wants to test that fork from 1.0.0; example response:

    
    <?xml version="1.0"?>
    <samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" ID="pfxd594347d-495f-b8d1-0ee2-41cfda14dd35" Version="2.0" IssueInstant="2015-01-02T22:48:48Z" Destination="http://localhost:9001/v1/users/authorize/saml" Consent="urn:oasis:names:tc:SAML:2.0:consent:unspecified" InResponseTo="_ed915a40-74fb-0132-5b16-48e0eb14a1c7">
      <Issuer xmlns="urn:oasis:names:tc:SAML:2.0:assertion">http://example.com</Issuer><ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
      <ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
        <ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
      <ds:Reference URI=""><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/><ds:DigestValue>jCgeXCPDFlwjTgqgPp0mU2Tqw9g=</ds:DigestValue></ds:Reference></ds:SignedInfo><ds:SignatureValue>Dfupr18wR+rDagtCPEfQlSGHzw4NdfPHZ4HszFe1JPCJXjfbyEM1fg+jzgGbMMt6XzgCXcKJM7E/CPSTDkvMe3DUJlHu4DhuDOAz/DsyoBwWuV+Rf3WiNcF4XCc9wBQzvn/atD7zW6xwO7N/hkAZJqfvJddnpMMHKGXqG/ZHZAw=</ds:SignatureValue>
[...]

PR #254, opened 2015-07-17
Handle empty URI references

knightlabs:
as per http://www.w3.org/TR/xmldsig-core/#sec-Same-Document; thx to @sixto for resolving a test case failure. Squashed down from knightlabs/ruby-saml/without_reference_uri


knightlabs:
See https://github.com/onelogin/ruby-saml/pull/253 for details of the original PR discussion.

knightlabs:
Thanks!

PR #255, opened 2015-07-21
Refactor validate signature. 

pitbulk:
Reorder code.
Avoid signature validation error when DS namespace declared at the root (don't remove signature element).


tcummings:
We are also running into the issue this resolves.  Any idea when it will be released to rubygems?

luisvm:
:+1: ship it

tcummings:
:+1:  sweetness!

PR #256, opened 2015-08-04
Separated the create_authentication_xml_doc in two methods

levelboy:
@pitbulk 
I ran in a problem that I can't find a good solution for.
I want to be able to add Scoping namespaces as per @tekt8tket 's  PR 

I do understand your need to keep the library simple 
I'm also thinking that the SAML standard is pretty big, and a library cannot cover all the usecases of SAML. 
Because of this, I find it imperative that the library should be open for extension.
The way Authrequest.create_authentication_xml_doc is now, with the signing of the document before it's returned, doesn't even allow easy monkey patching. 

So this is what I propose, a very simple split in the ````Authrequest.create_authentication_xml_doc```` between the creation and population of the document, and the signing, so you can hook your code between them.

What do you think?

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -87,6 +87,11 @@ module OneLogin
       # @return [String] The SAMLRequest String.
       #
       def create_authentication_xml_doc(settings)
+        document = create_xml_document(settings)
+        sign_document(document, settings)
+      end
+
+      def create_xml_document(settings)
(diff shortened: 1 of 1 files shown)


pitbulk:
I'm ok with this solution. Right now I'm working on implementing a new ruby-common-gem that will have the core functionalities (handle general SAML xmls, sign, validations, encryption, decryption) and the ruby-saml wil delegate on it all those functionality.
I will keep this in mind when releasing the new ruby-saml version.

levelboy:
Ok @pitbulk, thank you

PR #258, opened 2015-08-18
Fix allowed_clock_drift on the validate_session_expiration test

sthanson:
## Status
**READY**

## Migrations
NO

## Description
The allowed_clock_drift parameter was removing valid time from the test instead of padding it with extra time.

If assertion sets SessionNotOnOrAfter as 11:30am
and the test is padded with an allowed_clock_drift of 1 minute,
I would expect these results:

at 11:29 = test passes because it's before
at 11:30 = test passes because the allowed_clock_drift gave us an extra minute
at 11:31 = test fails because it's on the expiration time, so add an error
at 11:32 = test fails because it's after the expiration time, so add an error

## Related PRs
List related PRs against other branches:
none

## Todos
- [x] Tests
- [x] Documentation (no documentation necessary)

## Deploy Notes
This is a bug fix which requires no additional migration or settings change.

## Steps to Test or Reproduce
Run the validate_session_expiration check on an assertion that expires shortly and it should pass.  Now if you pad the same test with an allowed_clock_drift of 3000000 minutes, the test should still pass.  Before it was failing with a huge padding because the allowed_clock_drift removed valid time.

* The validate_session_expiration validation on the Response will be affected for anyone using the allowed_clock_drift option.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -515,7 +515,7 @@ module OneLogin
         return true if session_expires_at.nil?
 
         now = Time.now.utc
-        unless session_expires_at > (now + allowed_clock_drift)
+        unless (session_expires_at + allowed_clock_drift) > now
           error_msg = "The attributes have expired, based on the SessionNotOnOrAfter of the AttributeStatement of this Response"
           return append_error(error_msg)
         end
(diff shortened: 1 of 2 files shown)


pitbulk:
@sthanson  Thanks for contribute

PR #262, opened 2015-09-01
Add no_destination option

mathijssterk:
According to the SAML specification at http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf the Destination attribute is optional:

```
Destination [Optional] A URI reference indicating the address to which this request has been sent. This is useful to prevent malicious forwarding of requests to unintended recipients, a protection that is required by some protocol bindings. If it is present, the actual recipient MUST check that the URI reference identifies the location at which the message was received. If it does not, the request MUST be discarded. Some protocol bindings may require the use of this attribute (see [SAMLBind]).
```

Some of our customers do not handle the Destination attribute correctly (the Identity Provider keeps redirecting to itself), therefore we had to make it optional.

This pull-request adds a ```no_destination``` option to the settings to remove the attribute from the SAML Request.

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -101,7 +101,7 @@ module OneLogin
         root.attributes['ID'] = uuid
         root.attributes['IssueInstant'] = time
         root.attributes['Version'] = "2.0"
-        root.attributes['Destination'] = settings.idp_sso_target_url unless settings.idp_sso_target_url.nil?
+        root.attributes['Destination'] = settings.idp_sso_target_url unless settings.no_destination || settings.idp_sso_target_url.nil?
         root.attributes['IsPassive'] = settings.passive unless settings.passive.nil?
         root.attributes['ProtocolBinding'] = settings.protocol_binding unless settings.protocol_binding.nil?
         root.attributes["AttributeConsumingServiceIndex"] = settings.attributes_index unless settings.attributes_index.nil?
--- a/lib/onelogin/ruby-saml/settings.rb
+++ b/lib/onelogin/ruby-saml/settings.rb
@@ -48,6 +48,7 @@ module OneLogin
       attr_accessor :authn_context
       attr_accessor :authn_context_comparison
       attr_accessor :authn_context_decl_ref
+      attr_accessor :no_destination
       attr_reader :attribute_consuming_service
       # Work-flow
       attr_accessor :security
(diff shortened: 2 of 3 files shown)


pitbulk:
What Identity Providers experience that problem and why they not fix this issue?

mathijssterk:
I know one of our customers with this issue uses ADFS. It is probably something they could solve when they configure it correctly, but within large companies it tends to take ages (waiting on a response from them for more than a year).

My guess is that the idp_sso_target_url is referencing a proxy server that forwards the SAML Request to a local domain where ADFS resides. Then ADFS does not recognize the Destination attribute as being itself, as it is the domain name of the proxy server, and redirects there. Hence the loop.

As the Destination attribute is optional, they asked us to remove it. You could argue this is a workaround, but as it is allowed by the SAML 2.0 specs, I would very much appreciate it if this option could become part of the ruby-saml code.

pitbulk:
At https://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf when describing the HTTP-Redirect  Binding at page 19 (3 4.5 2 Security  Considerations). I read:
```
If the message is signed, the Destination XML attribute in the root SAML element of the protocol
message MUST contain the URL to which the sender has instructed the user agent to deliver the
message.
``` 

So I don't know if this "optional" parameter is required when the Message is signed.
Also I dislike the idea of adding an extra 'no_destination' option to the general settings. (And if added, why only affects AuthNRequest?)

mathijssterk:
I patched the AuthNRequest as that is the only one we use. I was not aware that the Destination attribute was used in other communication as well. The fact that the Destination attribute is required in a specific context does not rule out that it is optional in another context. We found that removing the Destination attribute alltogether did fix our communication with ADFS for some customers.

I agree that a ```no_destination``` option is dodgy. Maybe it is better to add a ```idp_sso_destination```  option that defaults to the ```idp_sso_target_url```? This makes it possible to distinguish between the URI of the proxy server and the domain name of actual IdP. And by explicitly setting the ```idp_sso_destination``` to an empty string the Destination attribute could be removed from the request.

What are you thoughts about that?

pitbulk:
Here is a note from ADFS.

http://social.technet.microsoft.com/wiki/contents/articles/3463.ad-fs-2-0-the-admin-event-log-contains-error-event-320-msis1010-signed-saml-message-must-have-destination-uri-specified.aspx

PR #263, opened 2015-09-09
Return only 'request_params'

rzuniga-gap:


pitbulk:
Can you explain this PR?

PR #271, opened 2015-10-09
Fixes #270 -- allow SAML elements to come from any namespace

### LINKED ISSUE #270: decrypt_assertion fails with XML from a SAML IDP which names XML namespaces differently
The code which parsers an assertion (https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L690-L692), uses regular expressions to match the XML, so it doesn't handle namespaces properly. Here is an example SAML assertion which it fails to parse:

```xml
<ns2:Assertion xmlns:ns2="urn:oasis:names:tc:SAML:2.0:assertion" ID="_d0a12c08368939434a522a5e716822d15a64" IssueInstant="2015-10-09T18:36:10Z" Version="2.0">
    ...
</ns2:Assertion>
```

alex:


paultag:
Any thoughts on this, @pitbulk?

Issue #275, opened 2015-10-29
Problem validating document with multiple Signature elements


levelboy:
The lastest release of the gem (1.1.0), fails to validate documents of this form:

````
<samlp:Response>
	<saml:Issuer></saml:Issuer>
	<ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
		<ds:SignedInfo>
			<ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
			<ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
			<ds:Reference URI="#_f7379ea04746ee86b0a16f519aea47c83cd0eefad0">
				<ds:Transforms>
					<ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
					<ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
				</ds:Transforms>
				<ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
				<ds:DigestValue></ds:DigestValue>
			</ds:Reference>
		</ds:SignedInfo>
		<ds:SignatureValue></ds:SignatureValue>
		<ds:KeyInfo>
			<ds:X509Data>
				<ds:X509Certificate></ds:X509Certificate>
			</ds:X509Data>
		</ds:KeyInfo>
	</ds:Signature>
	<samlp:Status>
		<samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/>
	</samlp:Status>
	<saml:Assertion ID="_a945e7e0e04ae098e9603cdcd9360f4a3d558198b1" IssueInstant="2014-12-08T16:22:43Z" Version="2.0" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
		<saml:Issuer></saml:Issuer>
		<ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
			<ds:SignedInfo>
				<ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
				<ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
[...]

PR #277, opened 2015-10-30
275 - Validates documents with signed message and assertion

levelboy:
# Status

READY

# Migrations

NO

# Description
see #275 
The modification is that when we check the signature of the response, we only check the first Reference object for that signature. According to the SAML standard, a signature node can only have one Reference node within.

# Related PRs

Problem seems to be introduced here: 275512a

# Todos

- Refactor xml_security to use only one XML library. At the moment it uses both nokogiri and REXML
- This only checks the signature on the document, as it was doing previously. It would be best and correct according to the SAML standard to verify each signatures in the document.

# Deploy Notes

No deploy notes. 

# Steps to Test or Reproduce

First commit adds relevant response document and failing test. Run ````rake test```` to see it failing.

### DIFF
--- a/lib/xml_security.rb
+++ b/lib/xml_security.rb
@@ -286,35 +286,34 @@ module XMLSecurity
       inclusive_namespaces = extract_inclusive_namespaces
 
       # check digests
-      REXML::XPath.each(sig_element, "//ds:Reference", {"ds"=>DSIG}) do |ref|
-        uri = ref.attributes.get_attribute("URI").value
+      ref = REXML::XPath.first(sig_element, "//ds:Reference", {"ds"=>DSIG})
+      uri = ref.attributes.get_attribute("URI").value
 
-        hashed_element = uri.empty? ? document : document.at_xpath("//*[@ID=$uri]", nil, { 'uri' => uri[1..-1] })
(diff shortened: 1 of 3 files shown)


levelboy:
@pitbulk Do you have any notes on this? 
Thank you.

pitbulk:
@levelboy Sorry, I had no time to review it yet.

pitbulk:
Hi @levelboy,

Thanks you very much for the PR, it seems that fixs the bug introduced at 1.1.0

I read again the SAML standard and you are right, SAML expects only 1 reference
```  
5.4.2 References
....
Signatures MUST contain a single <ds:Reference> containing a same-document reference to the ID
attribute value of the root element of the assertion or protocol message being signed. For example, if the ID attribute value is "foo", then the URI attribute in the <ds:Reference> element MUST be "#foo".
```

So we will apply your patch.

P.S We followed xmlsec lib steps that loop over all the references, but is true we don't need to do that:
https://github.com/robrichards/xmlseclibs/blob/1aaa0c20c9967d29a8cccdf81ad3994244c10d8d/src/XMLSecurityDSig.php#L572

levelboy:
Great, thanks @pitbulk

PR #289, opened 2015-11-26
Use secure random and encapsulate uuid creation

subfusc:
Rebase of and patched for backwards compatability for #206.

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -1,8 +1,8 @@
-require "uuid"
 require "rexml/document"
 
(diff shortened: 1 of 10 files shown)


pitbulk:
Thanks for contribute, I will review it and merge it as soon as possible ;)

subfusc:
Is there an ETA for this? As described in the linked topic, this is producing actual errors when using the library.

ilikepi:
A big :+1: to this.  We ran into an issue last night due to the way the uuid gem determines where to store its state file by default.

subfusc:
uuid is only included as a dependency if your ruby version is less than 1.9. At least that is the intent.

PR #292, opened 2015-12-03
Improve the error message when validating the audience

Jaco-Pretorius:
This change provides additional context when validating the audiences fails


BrentWheeldon:
:+1:

Issue #301, opened 2016-02-11
missing validation for Destination in response


alperkokmen:
it seems like destination validation never made it to `1.1.0`. it was [part of the original PR](https://github.com/onelogin/ruby-saml/pull/197/files#diff-58e0f05945ffec83ea2c44ce4e7b8550R262) @pitbulk opened that improved the validations among other things: #197 

however, it seems like `validate_destination` never made it to the smaller PRs created – i am guessing it should've been part of #239?

in the [spec](http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf), it's described on lines 1554-1559.

> Destination [Optional]
> A URI reference indicating the address to which this response has been sent. This is useful to prevent
> malicious forwarding of responses to unintended recipients, a protection that is required by some
> protocol bindings. If it is present, the actual recipient MUST check that the URI reference identifies the
> location at which the message was received. If it does not, the response MUST be discarded. Some
> protocol bindings may require the use of this attribute (see [SAMLBind]).

PR #239, opened 2015-06-01
Add more validations and refactor and improve tests.

pitbulk:
* Add more validations to the SAMLResponse class.
* Add more validations to the LogoutResponse class
* Add more validations to the LogoutRequest class
* Refactor and improve tests (+coverage).

PR #302, opened 2016-02-11
add Destination validation for ruby-saml/response

alperkokmen:
## Status
**READY**

## Migrations
NO

## Description
this change adds validation for `Destination` attribute which is optionally included in the SAML response.

- response method to return `Destination` attribute via `REXML`.
- response method to validate the destination against consumer service URL available in settings.
- unit tests.

note that @pitbulk originally implemented these changes in https://github.com/onelogin/ruby-saml/pull/197; however, this specific validation didn't make it to master branch (or >= 1.0.0). i merely duplicated his work and matched the new validation pattern (no `soft`) in `master`.

if accepted, this would fix https://github.com/onelogin/ruby-saml/issues/301.

## Related PRs

branch | PR
------ | ------
`response_validations`  | https://github.com/onelogin/ruby-saml/pull/197
`improve_validations` | https://github.com/onelogin/ruby-saml/pull/239

## Todos
- [x] Tests
- [x] Documentation – i don't think this is applicable; other validations aren't documented individually in `README`.

## Deploy Notes
no migrations or breaking changes; some IdPs may not send the right `Destination` due to configuration issues – since it's an _optional_ value, when provided, expectation is that it must be valid. to keep those integrations working, `#consume` action be tweaked to ignore validations for `Destination` attribute.

## Steps to Test or Reproduce
__N/A__

* Response validation.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -254,6 +254,19 @@ module OneLogin
         end
       end
 
+      # @return [String|nil] Destination attribute from the SAML Response.
+      #
+      def destination
+        @destination ||= begin
+          node = REXML::XPath.first(
+            document,
+            "/p:Response",
+            { "p" => PROTOCOL }
+          )
+          node.nil? ? nil : node.attributes['Destination']
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


pitbulk:
Thanks for the contribution.

PR #308, opened 2016-03-10
Adds Response#name_id_format

grzuy:


pitbulk:
I'm ok with this change, @luisvm  can you review it too?

PR #321, opened 2016-04-15
Expose IdP attribute names in SAML2 toolkit settings

alex-wood:
I've added a new attribute, `idp_attribute_names`, to the IdP section of the settings class. I've encountered a few scenarios where the ability to extract attribute names from the IdP metadata would be useful and figured others might be interested in such functionality as well.


pitbulk:
Can you add tests?

PR #324, opened 2016-04-27
Intermediate certificate support

alex-wood:
Adds support for intermediate certificates and corresponding test cases.

### DIFF
--- a/lib/onelogin/ruby-saml/metadata.rb
+++ b/lib/onelogin/ruby-saml/metadata.rb
@@ -39,6 +39,13 @@ module OneLogin
           add_sp_cert(sp_sso, "signing", sp_cert_text)
           add_sp_cert(sp_sso, "encryption", sp_cert_text)
 
+          intermediate_certs = settings.get_sp_intermediate_certs
+          if intermediate_certs
+            intermediate_certs.each do |intermediate_cert|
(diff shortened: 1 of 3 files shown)


pitbulk:
Why Intermediate certificate support is required? Can you explain?

alex-wood:
For certain certificates, there’s no path for verifying the trust chain unless one provides intermediate certificates in the metadata response. My changes follow the [spec for XML signatures](https://www.w3.org/TR/xmldsig-core/#sec-X509Data), which describes how intermediate signatures are included.

pitbulk:
In the toolkit, the Signature verification is done in 2 steps:
* Check that the signature value is actually corresponding to the content it protects
* Check that the signature was made by someone trusted by the verifier (the certificate matches the IdP certificate registered in the settings.

At the toolkit we don't verify that the certificate itself is valid, so I don't think that "intermediate certificate" stuff is required...

Can you provide a SAMLResponse not able to be validated?

alex-wood:
To be clear, the issue is about the IdP not accepting our metadata file because the trust chain for the service provider certificate is not there.

pitbulk:
I think this solves such an specific scenario and I dislike be adding new settings parameters for that.

In order to solve your case, I think you better extend the metadata class and modify the generate method to accept intermediate certs.

PR #326, opened 2016-05-20
ensure same validations are run whether collecting errors or not

s01ipsist:
`validate_destination` only occurred in one branch of the `validate` method. This change refactors to ensure the validations are defined only once and then called appropriately.


pitbulk:
@s01ipsist can you extend your PR and apply your solution at slo_logoutrequest.rb and logoutresponse.rb ?

s01ipsist:
@pitbulk done.

pitbulk:
Thanks for contribute!

Issue #327, opened 2016-05-27
multiple ConsumerService endpoints


lisovskyvlad:
So, looks like it is possible with saml  to have multiple AssertionConsumerService in metadata.

My case is that I would like to have one endpoint for testing purpose(I am Sp) and that IdP (shibboleth, sorry, not my choise) and another for production.

What do you think?

pitbulk:
Hi @lisovskyvlad 

Currently our toolkit only supports one AssertionConsumerService URL.

My suggestion: define 2 SP entities (you only need to dynamically change the setting data), so you will have 2 metadatas to be registered at the IdP.

Imagine you have prod.sp.com and test.sp.com running the same piece of code, in this case lets assume you are using the [ruby-saml-example](https://github.com/onelogin/ruby-saml-example).
The [get_url_base](https://github.com/onelogin/ruby-saml-example/blob/master/app/controllers/saml_controller.rb#L125) method will take care of the different url_base.
It will [build the  ruby-saml settings](https://github.com/onelogin/ruby-saml-example/blob/master/app/controllers/saml_controller.rb#L9) based on that value
```
settings = Account.get_saml_settings(get_url_base)
```

[SP data will be built dinamically](https://github.com/onelogin/ruby-saml-example/blob/master/app/models/account.rb#L14) but [IdP data](https://github.com/onelogin/ruby-saml-example/blob/master/app/models/account.rb#L19) will be the same (Shibboleth IdP metadata).

lisovskyvlad:
Hey @pitbulk,

thanks for quick answer, yeah, I have this approach with dynamic metadata (dep. of host) generating, but looks like IdP is not so flexible (but I am not sure) and they (IT consultant of if;))) have to chane this settings because of flexibility.

Anyway, thanks. What is about patch (I could try to write one) for this possibility? The goal is that metadata has possibility to have few ACS node. 
May be it is not needed for onelogin IdP service, but it fulfils saml 2.0 spec, so.

pitbulk:
At the IdP-initiaited SSO flow, always the default ACS will be used
At the Sp-initiated SSO flow you will need to inform to the  IdP what ACS should be used (set a AssertionConsumerServiceURL parameter of the AuthNRequest).

Write a patch to support that is not really complex, but will increase the complexity of the toolkit.
[...]

PR #328, opened 2016-05-27
Added support for AssertionConsumerServiceIndex

joshpfosi:
## Status
**READY**

## Migrations
NO

## Description
This pull request adds support for the `AssertionConsumerServiceIndex` to allow
for multiple development environments to work with an inflexible IdP.

## Todos
- I am unable to test this in production as I am using `ruby-saml` in a gem
which incorporate into many other apps. As Rails does not allow gem dependencies
to be local paths or Git/Github repositories, I need this to be merged to fully
test it. This is a very simple PR, however, and so its correctness can likely be
ascertained by inspection.

## Deploy Notes
N/A

* Orthogonal to current features.

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -111,6 +111,9 @@ module OneLogin
         if settings.assertion_consumer_service_url != nil
           root.attributes["AssertionConsumerServiceURL"] = settings.assertion_consumer_service_url
         end
+        if settings.assertion_consumer_service_index != nil
+          root.attributes["AssertionConsumerServiceIndex"] = settings.assertion_consumer_service_index
+        end
         if settings.issuer != nil
           issuer = root.add_element "saml:Issuer"
           issuer.text = settings.issuer
--- a/lib/onelogin/ruby-saml/metadata.rb
+++ b/lib/onelogin/ruby-saml/metadata.rb
@@ -69,7 +69,7 @@ module OneLogin
               "Binding" => settings.assertion_consumer_service_binding,
               "Location" => settings.assertion_consumer_service_url,
               "isDefault" => true,
-              "index" => 0
+              "index" => settings.assertion_consumer_service_index
           }
         end
 
--- a/lib/onelogin/ruby-saml/settings.rb
+++ b/lib/onelogin/ruby-saml/settings.rb
@@ -33,6 +33,7 @@ module OneLogin
       attr_accessor :issuer
       attr_accessor :assertion_consumer_service_url
       attr_accessor :assertion_consumer_service_binding
+      attr_accessor :assertion_consumer_service_index
       attr_accessor :sp_name_qualifier
       attr_accessor :name_identifier_format
       attr_accessor :name_identifier_value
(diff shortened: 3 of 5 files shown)


joshpfosi:
This addresses #327.

pitbulk:
What the point of added index support if only 1 ACS URL is described on the settings? I'm lost.

BTW. In order to test ruby-saml on [ruby-saml-example](https://github.com/onelogin/ruby-saml-example) I used to define at the Gemfile
```
gem 'ruby-saml', :path => '/home/projects/ruby-saml
```
and worked.

joshpfosi:
Right, using the path works in a given project. What does not work is using the path in a gem which is then used in that given project. I actually worked around this by building the gem and deploying it onto our gem server.

I am very new to SAML so perhaps I have misunderstood. My IT administrator made it seem as if this was common practice. Namely, we have multiple environments in which we test our application e.g. http://{app-dev, app-staging, app-beta, app}.net/ if you will. Rather than generate and maintain metadata for each of those apps (our IT department was unwilling to do this as they have hundreds of apps w/ SSO), they suggested we use the same entity ID, generate multiple metadata sets each with unique ACS URL and Index. The index (apparently) will be used by the IdP to choose which URL to use.

To summarize: Each of our environments would submit its own "version" of the metadata with an ACS URL and Index. The IdP, however, would have a merged set of metadata with a single map between index and URL.

pitbulk:
Yes, and this is what we discussed on #327.

In the SAML standard you are able to define multiple ACS URLS, but it used to work to define different binding endpoints, for example, this can be found at a SP metadata:

```
 <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://sp.example.com/simplesaml/module.php/saml/sp/saml2-acs.php/default-sp" index="0"/>
    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:1.0:profiles:browser-post" Location="https://sp.example.com/simplesaml/module.php/saml/sp/saml1-acs.php/default-sp" index="1"/>
    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Artifact" Location="https://sp.example.com/simplesaml/module.php/saml/sp/saml2-acs.php/default-sp" index="2"/>
    <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:1.0:profiles:artifact-01" Location="https://sp.example.com/simplesaml/module.php/saml/sp/saml1-acs.php/default-sp/artifact" index="3"/>
```

But this toolkit only uses for the AssertionConsumerService endpoint the HTTP-POST binding, so at the ruby settings we only let to define one unique ACS URL and we assume that its index is 1.

I understand the "trick" that your IT department suggested to you, but as I already told @joshpfosi, support that will add complexity to the current gem:
- Be able to define on the settings more than 1 ACS URL and keep backward compatibility
- Extend login method to specify the ACS_UR/ACS_index to be used on the AuthNRquest
- Extend AuthNRquest constructor to support the ACS_UR/ACS_index desired and build the XML with the right AssertionConsumerServiceURL
- Modify the SAMLResponse validator to cares about multiple ACS_URLS

As you can see multiple changes and complexity for covering your specific case.
That is why I suggested at #327 to extend the classes and methods for your specific scenario.

joshpfosi:
Ahh I see. Thank you for clarifying that, I understand it much better now. Apologies for the inconvenience.

Issue #330, opened 2016-05-30
IdpMetadataParser#parse method says it receives an Hash, but uses it as a Settings object


sanderhahn:
`IdpMetadataParser#parse` method says it receives an Hash, however the code seems to use the options as an `Settings` object: https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/idp_metadata_parser.rb#L45
Changing the code into: `OneLogin::RubySaml::Settings.new(options[:settings] || {}).tap do |settings|` would make the code work like the @param says.

sanderhahn:
Might have been misreading the docs, this maybe a non-issue.

pitbulk:
Solved at:
- [Allow multiple authn_context_decl_ref in settings](https://github.com/onelogin/ruby-saml/commit/63e4b7a4e342fe9fb541a2cba634c718ed44f85f)
- [Allow options[:settings] to be an hash for Settings overrides in IdpMetadataParser#parse](https://github.com/onelogin/ruby-saml/commit/4616771ee06b2f3b42c3d74b4b23755b5d2f84cc)
- [include vendor folder on SimpleCov filters](https://github.com/onelogin/ruby-saml/commit/4fb48d692567d6391b6e326989edb3a486978c6b)

PR #331, opened 2016-05-31
Use settings as hash for Settings overrides in IdpMetadataParser#parse

sanderhahn:
Intents to fix issue #330

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -141,8 +141,14 @@ module OneLogin
           end
           # add saml:AuthnContextDeclRef element
           if settings.authn_context_decl_ref != nil
-            class_ref = requested_context.add_element "saml:AuthnContextDeclRef"
-            class_ref.text = settings.authn_context_decl_ref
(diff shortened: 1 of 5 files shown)


sanderhahn:
Travis CI seems to have failed on jruby on unrelated code.

pitbulk:
It was a temp problem with the CI.

What do you think if we check the value of options[:settings] (if is provided in the method) and then based on what kind of object is, we call or not the  OneLogin::RubySaml::Settings constructor?

pitbulk:
Solved at:
- [Allow multiple authn_context_decl_ref in settings](https://github.com/onelogin/ruby-saml/commit/63e4b7a4e342fe9fb541a2cba634c718ed44f85f)
- [Allow options[:settings] to be an hash for Settings overrides in IdpMetadataParser#parse](https://github.com/onelogin/ruby-saml/commit/4616771ee06b2f3b42c3d74b4b23755b5d2f84cc)
- [include vendor folder on SimpleCov filters](https://github.com/onelogin/ruby-saml/commit/4fb48d692567d6391b6e326989edb3a486978c6b)

PR #340, opened 2016-07-12
More tests for XML Signature Wrapping

tgoddard:
This includes some passing tests for XML signature wrapping attacks. The library appears to handle these well at the moment, but since I wrote the tests to confirm this, thought I'd submit them to guard against regression.


pitbulk:
Thanks, I will review them and add to the test suite.

PR #344, opened 2016-07-20
Add Response#errors with an attr_reader

hrs:
Being able to access the list of errors on a response is sometimes convenient for debugging in development. This exposes the `@errors` variable through a public method.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -25,6 +25,7 @@ module OneLogin
 
       attr_reader :document
       attr_reader :decrypted_document
+      attr_reader :errors
       attr_reader :response
       attr_reader :options
 


pitbulk:
@hrs 

The attr_accessor :errors was defined at the [ErrorHandling](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/error_handling.rb#L6) class, used at [Response](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L14) , [SloLogoutrequest](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/slo_logoutrequest.rb#L14) and [Logoutresponse](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/logoutresponse.rb#L13).

hrs:
Oh, whoops, sorry for the noise @pitbulk! We didn't see it as at `attr_reader`, assumed it was missing, and monkey-patched the class in development to add it (which was, apparently, totally unnecessary).

Thanks for the pointer, and for the great gem!

pitbulk:
no problem ;)

PR #345, opened 2016-07-28
Support multiple settings.auth_context

pkarman:
**Why**: SAML supports multiple AuthnContextClassRef elements
in an Authn request.


pitbulk:
@luisvm can you review and give a :+1: ?

pitbulk:
@pkarman can you add to the README a reference about this parameter?
also rather than "foo" and "bar", can you use real AuthnContextClassRef values?

pkarman:
@pitbulk done -- let me know if you'd like the commits squashed.

luisvm:
lgtm 👍

PR #347, opened 2016-08-04
Don't recreate the attributes object when there are no attributes

gregbeech:
Because of the early `return` this method was creating a new attributes
object each time it was called. When this is used in omniauth-saml
which sets `response.attributes[‘fingerprint’] =
options.idp_cert_fingerprint` the fingerprint is lost because the
object is discarded — this is problematic when you need to access the
fingerprint to see which IdP sent the SAML because the claims might not
apply to the entire entity-space; only those ‘owned’ by a particular
IdP.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -32,7 +32,7 @@ module OneLogin
 
       # Constructs the SAML Response. A Response Object that is an extension of the SamlMessage class.
       # @param response [String] A UUEncoded SAML response from the IdP.
-      # @param options  [Hash]   :settings to provide the OneLogin::RubySaml::Settings object 
+      # @param options  [Hash]   :settings to provide the OneLogin::RubySaml::Settings object
       #                          Or some options for the response validation process like skip the conditions validation
       #                          with the :skip_conditions, or allow a clock_drift when checking dates with :allowed_clock_drift
       #                          or :matches_request_id that will validate that the response matches the ID of the request,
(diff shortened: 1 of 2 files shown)


gregbeech:
Looks like the jruby build failed with an error, but it seems unrelated to this change:

```
  1) Error:
Authrequest#test_0002_create the deflated SAMLRequest URL parameter including the Destination:
Zlib::BufError: buffer error
    org/jruby/ext/zlib/ZStream.java:134:in `finish'
    /home/travis/build/onelogin/ruby-saml/test/request_test.rb:35:in `test_0002_create the deflated SAMLRequest URL parameter including the Destination'
```

pitbulk:
response.`attributes` must return the attributes founded at the SAMLResponse, you should not store anything there.

I don't know why at [omniauth-saml](https://github.com/omniauth/omniauth-saml/blob/master/lib/omniauth/strategies/saml.rb#L71) you are storing this 
```
response = OneLogin::RubySaml::Response.new(request.params['SAMLResponse'], opts.merge(settings: settings))
response.attributes['fingerprint'] = options.idp_cert_fingerprint
```

but if you want to inject data in the userdata set, why don't do it a[fter retrieve the SAML Response attributes](https://github.com/omniauth/omniauth-saml/blob/master/lib/omniauth/strategies/saml.rb#L78)?

```
@name_id = response.name_id
@attributes = response.attributes
@attributes['fingerprint'] = options.idp_cert_fingerprint
@response_object = response
```

gregbeech:
Hi @pitbulk I didn't write the `omniauth-saml` code, am just looking to use it for supporting SAML in our codebase and need to be able to determine the IdP which isn't currently possible without some way of looking it up. As they already set the fingerprint (just not reliably because attributes is recreated every time if it was originally empty) this was just an attempt to fix their code at the root.

But I can see your point, storing things that didn't come from the response in the attributes is a bit iffy, so perhaps a better approach would be to submit a PR to `omniauth-saml` to allow flowing additional metadata through in the `extra` part of the login hash so it's possible to determine which IdP provided the claims.

pitbulk:
by the way, I checked omniauth-saml's settings/code and I don't understand the use of
```
 :idp_cert_fingerprint_validator     => lambda { |fingerprint| fingerprint },
```

At the ruby toolkit, in order to check embedded Signatures (of the HTTP-POST binding), when you add a `:idp_cert_fingerprint` instead the `:idp_cert`, doesn't matter, at the end the idp_cert is [turned](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/settings.rb#L107) in a [idp_cert_fingerprint](https://github.com/onelogin/ruby-saml/blob/master/lib/xml_security.rb#L200) to [validate](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L683) the document.

The certificate of the SAMLResponse is [fingerprinted](https://github.com/onelogin/ruby-saml/blob/master/lib/xml_security.rb#L222) and [compared](https://github.com/onelogin/ruby-saml/blob/master/lib/xml_security.rb#L224) with the value of the idp_cert_fingerprint.

I always recommend to set the `idp_cert` and not the `idp_cert_fingerprint` because HTTP-Redirect binding signature validations requires it (since the IdP's public certificate is not at the SAML Message).

pitbulk:
yeah, maybe create that PR there, and reference this one.

PR #354, opened 2016-09-23
Allows scheme and domain to match ignoring case

tdphillipsjr:
- Per RFC4343, the domain name portion of a URI should be considered case-insensitive.
- Per RFC3986, the scheme portion of a URI should be considered case-insenstive.
- Some SSO providers allow users to enter their own subdomain, which many may do with capital letters (such as in the case of an acronym).
- The destination match should take this in to consideration when matching the destination URI to the ACS URI, if these are proper URIs.
- The match should default to the original case when either of the values are not proper URIs.

### DIFF
--- a/lib/onelogin/ruby-saml/logoutresponse.rb
+++ b/lib/onelogin/ruby-saml/logoutresponse.rb
@@ -195,7 +195,7 @@ module OneLogin
       def valid_issuer?
         return true if settings.idp_entity_id.nil? || issuer.nil?
 
-        unless URI.parse(issuer) == URI.parse(settings.idp_entity_id)
+        unless OneLogin::RubySaml::Utils.uri_match?(issuer, settings.idp_entity_id)
           return append_error("Doesn't match the issuer, expected: <#{settings.idp_entity_id}>, but was: <#{issuer}>")
         end
         true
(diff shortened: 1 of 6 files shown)


tdphillipsjr:
Hi @pitbulk -- I was hoping you guys could take a look at this PR and decide if it can be incorporated.  The domain case-insensitivity issue is blocking us from upgrading this Gem and resolving a CVE.  We're going to have to switch to a local fork if this or something similar can't be done in this gem, which is something we'd really like to avoid.

Thanks.

pitbulk:
Sorry for the delay.

May we use that solution to check also the issuers?

https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L568
https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/slo_logoutrequest.rb#L212
https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/logoutresponse.rb#L195

Then maybe move the uri_match? and original_uri_match? to the [utils.py](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/utils.rb)?

tdphillipsjr:
@pitbulk Sure, that sounds good.  Is this something you'd like me to add to the PR?

pitbulk:
If you can yes, please!.

tdphillipsjr:
Hi @pitbulk -- this is done, as requested.

tdphillipsjr:
Hi @pitbulk -- just checking in here.

PR #363, opened 2016-11-14
Allow for multiple RequestedAttribute AttributeValues

### LINKED ISSUE #362: Support multiple attribute values per attribute
In an example SAML AttributeStatement found [here](https://en.wikipedia.org/wiki/SAML_2.0#SAML_2.0_Assertions), the point of providing AttributeValue is to provide the IDP options for input in the associated attribute.

This gem does not currently support supplying multiple AttributeValue elements for a given Attribute.  Doesn't this actually mean the IDP could only provide a single value for this attribute, mostly rendering it useless?

```xml
   <saml:AttributeStatement>
     <saml:Attribute
       xmlns:x500="urn:oasis:names:tc:SAML:2.0:profiles:attribute:X500"
       x500:Encoding="LDAP"
       NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri"
       Name="urn:oid:1.3.6.1.4.1.5923.1.1.1.1"
       FriendlyName="eduPersonAffiliation">
       <saml:AttributeValue
         xsi:type="xs:string">member</saml:AttributeValue>
       <saml:AttributeValue
         xsi:type="xs:string">staff</saml:AttributeValue>
     </saml:Attribute>
   </saml:AttributeStatement>
```

Is there currently a way to accept multiple values per attribute? if not, would there be interest in a pull request to allow the `attribute_value` property to accept an array of values?

For example...

```ruby
      settings.attributes_index = 0
      settings.attribute_consuming_service.configure do
        service_name "Service"
        service_index 0
        add_attribute(
          name: "urn:oid:1.3.6.1.4.1.5923.1.1.1.1",
          name_format: "urn:oasis:names:tc:SAML:2.0:attrname-format:uri",
          friendly_name: "eduPersonAffiliation"
          attribute_value: ["member", "staff"]
        )
      end
```

jondavidjohn:
Closes #362 

This allows the service provider to specify multiple acceptable values for an attribute within an attribute consumer service.

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,3 +614,5 @@ settings.attribute_consuming_service.configure do
   add_attribute :name => "Another Attribute", :name_format => "Name Format", :friendly_name => "Friendly Name", :attribute_value => "Attribute Value"
 end
 ```
+
+The `attribute_value` option additionally accepts an array of possible values.
--- a/lib/onelogin/ruby-saml/metadata.rb
+++ b/lib/onelogin/ruby-saml/metadata.rb
@@ -89,8 +89,10 @@ module OneLogin
               "FriendlyName" => attribute[:friendly_name]
             }
             unless attribute[:attribute_value].nil?
-              sp_attr_val = sp_req_attr.add_element "saml:AttributeValue"
-              sp_attr_val.text = attribute[:attribute_value]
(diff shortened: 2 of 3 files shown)


pitbulk:
thanks for contributing

PR #367, opened 2017-01-04
[ENT-38] Add separate :want_assertions_encrypted flag to settings.security

irvingreid:
As part of fixing ENT-38, I'd like to add a feature to OneLogin ruby_saml to have separate control for publishing signed authentication requests vs. requesting encrypted assertions. That will allow us to turn off encrypted assertions until we can fix the underlying bugs.

### DIFF
--- a/lib/onelogin/ruby-saml/metadata.rb
+++ b/lib/onelogin/ruby-saml/metadata.rb
@@ -36,17 +36,23 @@ module OneLogin
         cert = settings.get_sp_cert
         if cert
           cert_text = Base64.encode64(cert.to_der).gsub("\n", '')
-          kd = sp_sso.add_element "md:KeyDescriptor", { "use" => "signing" }
-          ki = kd.add_element "ds:KeyInfo", {"xmlns:ds" => "http://www.w3.org/2000/09/xmldsig#"}
-          xd = ki.add_element "ds:X509Data"
-          xc = xd.add_element "ds:X509Certificate"
-          xc.text = cert_text
 
-          kd2 = sp_sso.add_element "md:KeyDescriptor", { "use" => "encryption" }
-          ki2 = kd2.add_element "ds:KeyInfo", {"xmlns:ds" => "http://www.w3.org/2000/09/xmldsig#"}
-          xd2 = ki2.add_element "ds:X509Data"
-          xc2 = xd2.add_element "ds:X509Certificate"
-          xc2.text = cert_text
+          if settings.security[:authn_requests_signed]
+            cert_text = Base64.encode64(cert.to_der).gsub("\n", '')
+            kd = sp_sso.add_element "md:KeyDescriptor", { "use" => "signing" }
+            ki = kd.add_element "ds:KeyInfo", {"xmlns:ds" => "http://www.w3.org/2000/09/xmldsig#"}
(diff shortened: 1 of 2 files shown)


irvingreid:
Oops - meant to put this pull request against my fork and get it reviewed locally first. Happy to have feedback from you on it in any case.

pitbulk:
I think it makes sense, we are currently supporting it on other [SAML toolkits](https://github.com/onelogin/php-saml/blob/master/advanced_settings_example.php#L52)
There we also have the wantNameIdEncrypted so if it or the wantAssertionsEncrypted are active, [we publish the X509KeyDescriptor for encription](https://github.com/onelogin/php-saml/blob/af4126b34cca8608b47bb89b4f19ce8547464ba0/lib/Saml2/Settings.php#L779).

lib/onelogin/ruby-saml/metadata.rb
```diff
@@ -36,17 +36,23 @@ def generate(settings, pretty_print=false)
...
-          xd = ki.add_element "ds:X509Data"
-          xc = xd.add_element "ds:X509Certificate"
-          xc.text = cert_text
 
-          kd2 = sp_sso.add_element "md:KeyDescriptor", { "use" => "encryption" }
-          ki2 = kd2.add_element "ds:KeyInfo", {"xmlns:ds" => "http://www.w3.org/2000/09/xmldsig#"}
-          xd2 = ki2.add_element "ds:X509Data"
-          xc2 = xd2.add_element "ds:X509Certificate"
-          xc2.text = cert_text
+          if settings.security[:authn_requests_signed]
```
pitbulk: We need to provide md:KeyDescriptor["use"="signing"] not only when 
:authn_requests_signed is true, instead also when 
:logout_requests_signed or :logout_responses_signed, so please
update the condition and add the related tests.

irvingreid:
I took the approach of always including the certificate for signing if it is available, as the existing code did, rather than making it conditional on any of the specific signing flags.

pitbulk:
Thanks for contribute

j8th:
I am currently using ruby-saml 1.4.2.

I am researching the `:want_assertions_encrypted` flag, and I came across this PR as well as PR #369.  I am trying to figure out what this means for me.

I have not changed the `:want_assertions_encrypted` flag, so it is the default value of `false`, so my metadata does not include a `KeyDescriptor use="encryption"` node in the XML.

Based on what I've read, it sounds like if I set this flag to `true` to get the encryption `KeyDescriptor` in my metadata, then it is possible that some sort of failure could occur with my existing SSO.  I think it sounds like the issue could occur if one of the X509 certificates involved has a very large serial number (`X509SerialNumber`).

It sounds like if I do not want to risk this failure, then I need to upgrade to at least version 1.4.3.  Then I could turn on `:want_assertions_encrypted` without worry.

Any help would be appreciated, @pitbulk and @irvingreid.  Thanks!

irvingreid:
@j8th if your server's SAML certificate (the one you configure in `settings.certificate`) doesn't have a very large serial number, it's safe to turn on `:want_assertions_encrypted` with ruby-saml 1.4.2. Otherwise, yes, you should upgrade to 1.4.3 first.

j8th:
@irvingreid Thank you, that clears things up!  And thanks for the speedy response.

PR #369, opened 2017-01-06
Validate serial number as string to work around libxml2 limitation

irvingreid:
Feel free to toss this pull request if you don't think it belongs in the main product; we can keep it in a private fork if necessary.

Based on discussion in https://github.com/benoist/xmldsig/issues/31 and https://mail.gnome.org/archives/xml/2008-March/thread.html#00044, I agree with the suggestion in https://github.com/onelogin/ruby-saml/compare/master...mokhan:x509serialnumber?expand=1 to hack the schema for certificate serial number validation in order to work around libxml2's arbitrary restriction on the size of XML Schema integers.

I have tested this against ADFS 2016 and our Ruby SP was able to process an encrypted assertion.

### DIFF
--- a/lib/schemas/xmldsig-core-schema.xsd
+++ b/lib/schemas/xmldsig-core-schema.xsd
@@ -188,7 +188,7 @@
 <complexType name="X509IssuerSerialType"> 
   <sequence> 
     <element name="X509IssuerName" type="string"/> 
-    <element name="X509SerialNumber" type="integer"/> 
+    <element name="X509SerialNumber" type="string"/>
   </sequence>
 </complexType>
 


irvingreid:
Good heavens that's twice I've made the same mistake. Again, this was supposed to be opened against our fork. (/me shakes fist at Github UI)

pitbulk:
instead of 'string' can you check if xs:hexBinary works on your case?

irvingreid:
I expect `xs:hexBinary` won't work correctly, because the serial number is expressed in the XML document in decimal and we don't want libxml2 to interpret it as hex.

mokhan:
👍

mokhan:
Is it possible to push out a new version of the gem with this change, please?

pitbulk:
I will release a new version next week with that and other improvements.

mokhan:
Thanks @pitbulk!

PR #379, opened 2017-03-21
Allow no Audience in schema

heironimus:
I have an assertion with element `<saml:AudienceRestriction/>` which gives the following error:
```
OneLogin::RubySaml::ValidationError: Element '{urn:oasis:names:tc:SAML:2.0:assertion}AudienceRestriction': Missing child element(s). Expected is ( {urn:oasis:names:tc:SAML:2.0:assertion}Audience ).
```

According to section 2.5.1 of the [SAML spec](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf), this appears to be valid.

Modifying the schema fixes this.

### DIFF
--- a/lib/schemas/saml-schema-assertion-2.0.xsd
+++ b/lib/schemas/saml-schema-assertion-2.0.xsd
@@ -142,7 +142,7 @@
         <complexContent>
             <extension base="saml:ConditionAbstractType">
                 <sequence>
-                    <element ref="saml:Audience" maxOccurs="unbounded"/>
+                    <element ref="saml:Audience" minOccurs="0" maxOccurs="unbounded"/>
                 </sequence>
             </extension>
         </complexContent>


pitbulk:
At the SAML spec that you mentioned, at 2.5.1.4 Elements <AudienceRestriction> and <Audience>
you will find the following xsd part:
```
<element name="AudienceRestriction" type="saml:AudienceRestrictionType"/>
  <complexType name="AudienceRestrictionType">
    <complexContent>
      <extension base="saml:ConditionAbstractType">
        <sequence>
           <element ref="saml:Audience" maxOccurs="unbounded"/>
        </sequence>
      </extension>
    </complexContent>
  </complexType>
<element name="Audience" type="anyURI"/>
```
You can see that Audience is defined as:
```
<element ref="saml:Audience" maxOccurs="unbounded"/>
```
and not as
```
<element ref="saml:Audience" minOccurs="0" maxOccurs="unbounded"/>
```

At least 1 Audience element is required. See also [saml:AudienceRestrictionType](http://www.datypic.com/sc/saml2/t-saml_AudienceRestrictionType.html)

heironimus:
Yep. I see that now. I had completely looked past that.  Thanks.

PR #380, opened 2017-03-21
Fix binding configuration example in README.md

tonycchung:
the setting in the readme should be `single_logout_service_binding`, not `assertion_consumer_logout_service_binding`.

Ran into this small issue while needing to configure SLO with the http post binding :)


pitbulk:
Ty

PR #391, opened 2017-04-20
Attempt to detect binding for SSO and SLO

jondavidjohn:
We've got some metadata that looks like this...

```
<md:SingleSignOnService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://......"/>
```

The parser is trying to find

```
<md:SingleSignOnService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" Location="https://......"/>
```

and failing because of the defaulting to `HTTP-Redirect`.

These changes allow the parser to find the correct SSO url because it will attempt to parse the binding instead of falling back to "HTTP-Redirect"

### DIFF
--- a/lib/onelogin/ruby-saml/idp_metadata_parser.rb
+++ b/lib/onelogin/ruby-saml/idp_metadata_parser.rb
@@ -142,7 +142,7 @@ module OneLogin
       # @return [String|nil] SingleSignOnService endpoint if exists
       #
       def single_signon_service_url(options = {})
-        binding = options[:sso_binding] || "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"
+        binding = options[:sso_binding] || single_signon_service_binding || "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"
         node = REXML::XPath.first(
           document,
           "/md:EntityDescriptor/md:IDPSSODescriptor/md:SingleSignOnService[@Binding=\"#{binding}\"]/@Location",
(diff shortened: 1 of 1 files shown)


pitbulk:
but why you don't use the options[:slo_binding] to give priority?

jondavidjohn:
@pitbulk I don't think I would know that at run-time.  Assuming we're supporting multiple IDPs, this will likely change per IDP right?

To try and be clear, in order to use the priority, I'd need to provide it here...

```ruby
      idp_metadata_parser = OneLogin::RubySaml::IdpMetadataParser.new
      idp_metadata_parser.parse(file.read, { sso_binding: "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"})
```

The problem is that the IDP is specifying the binding in the metadata, so I don't know it at this point.

Is there a reason we don't want to allow the parser to try and parse the correct binding?

pitbulk:
@jondavidjohn I think I got it.

I'm right now reviewing the idp metadata parser so I will re-think about the current implementation..

I think the problem here is that if the priority binding is not found on the metadata, then nothing is returned instead of return the endpoint with the binding.

pitbulk:
@jondavidjohn thanks! sorry but I misunderstood your PR, it makes totally sense

pitbulk:
I merged it, but I will review and try to find better solution..
right now for me the code makes no sense..
```
binding = options[:sso_binding] || single_signon_service_binding || "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"
```

if single_signon_service_binding returns nothing means that no endpoint was found.. so no sense to set 
it as "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" and try to search again...
so something like:
```
      def single_signon_service_url(options = {})
        binding = single_signon_service_binding(options[:sso_binding])
        unless binding.nil?
          node = REXML::XPath.first(
            document,
            "/md:EntityDescriptor/md:IDPSSODescriptor/md:SingleSignOnService[@Binding=\"#{binding}\"]/@Location",
            { "md" => METADATA }
          )
          node.value if node
        end
      end
```

PR #393, opened 2017-05-10
Implement IdpMetadataParser#parse_to_hash

tosch:
…and IdpMetadataParser#parse_remote_to_hash.

Having the parsed metadata as Hash may be useful for configuring omniauth-saml, for instance. See omniauth/omniauth-saml#135

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ And on 'signing' and 'encryption' arrays, add the different IdP x509 public cert
 The method above requires a little extra work to manually specify attributes about the IdP.  (And your SP application)  There's an easier method -- use a metadata exchange.  Metadata is just an XML file that defines the capabilities of both the IdP and the SP application.  It also contains the X.509 public
 key certificates which add to the trusted relationship.  The IdP administrator can also configure custom settings for an SP based on the metadata.
 
-Using ```idp_metadata_parser.parse_remote``` IdP metadata will be added to the settings withouth further ado.
+Using ```idp_metadata_parser.parse_remote``` IdP metadata will be added to the settings without further ado.
 
 ```ruby
 def saml_settings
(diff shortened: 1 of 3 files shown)


tosch:
I briefly considered changing the interface of the IdpMetadataParser even further:

```ruby
parser = OneLogin::RubySaml::IdpMetadataParser::FromString.new(xml_string, parse_options)
# or
parser = OneLogin::RubySaml::IdpMetadataParser::FromURL.new(url, parse_options)
# or maybe even
parser = OneLogin::RubySaml::IdpMetadataParser::FromFile.new(url, parse_options)

# maybe with some helper class methods:
parser = OneLogin::RubySaml::IdpMetadataParser.new_for_string(xml_string, parse_options)
# etc.

parser.to_settings # returns OneLogin::RubySaml::Settings
parser.to_hash # returns an Hash
```

while this would be an imho cleaner interface (and avoids the side effects of setting instance variables in the current `#parse` method), I did not implement that as it would break the current API. If you want to see that in a future major release — I'm happy to contribute that.

pitbulk:
Nice work,

related to the rename that you mentioned [here](https://github.com/onelogin/ruby-saml/pull/393#issuecomment-300589556), rest of toolkits already uses the same name convention, so at the end is complex to apply a change like that.

Reviewing your PR, I noticed that we renamed options by parse_options.

I think we need to rename it back to options, since it can contains parse_options, but also a settings object.
And update the parameter description that currently only mention the [settings stuff](https://github.com/onelogin/ruby-saml/blob/key_rollover_mngmt/lib/onelogin/ruby-saml/idp_metadata_parser.rb#L31):

> @param options  [Hash]   :settings to provide the OneLogin::RubySaml::Settings object or an hash for Settings overrides

What do you think?

tosch:
@pitbulk I've updated the PR accordingly. I was not sure about all the `@option`s, so please review those descriptions carefully.

pitbulk:
I think now is perfect

PR #396, opened 2017-05-15
fixed issue with test/response_test.rb

geneshuman:
## Status
**READY**

## Migrations
NO

## Description
It looks like there is an issue in one of the test files.  Not entirely certain how it hasn't been noticed. basically let(:settings) isn't in scope all of the places it needs to be.  Very simple fix.

## Todos
- [X] Tests

## Steps to Test or Reproduce
`ruby test/response_test.rb`
in a fresh copy of the repo raises a bunch of errors

Test suite


geneshuman:
Not really sure why my change would be causing whatever error that is.

pitbulk:
Thanks for contribute.

PR #397, opened 2017-05-16
added SubjectConfirmation Recipient validation

geneshuman:
## Status
**READY**

## Migrations
NO

## Description
The SAML specificiation optionally specifies a Recipient field in the SubjectConfirmation element.  I believe this tag is not currently being validated.  See [here](https://www.google.com/search?q=saml+destination+vs+recipient&rlz=1C5CHFA_enUS721US721&oq=saml+destination+vs+recipient&aqs=chrome..69i57.4497j0j7&sourceid=chrome&ie=UTF-8).   I believe this correlates with settings.assertion_consumer_service_url, so I've added this check.

## Related PRs
NA

## Todos
- [X] Tests

## Deploy Notes
N/A

* More accurate adherence to the SAML spec, additional security.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -41,6 +41,12 @@ module OneLogin
         raise ArgumentError.new("Response cannot be nil") if response.nil?
 
         @errors = []
+
+        # skip recipient check by default for backwards compatibility
+        unless options.key?(:skip_recipient_check)
+          options[:skip_recipient_check] = true
+        end
+
         @options = options
         @soft = true
         unless options[:settings].nil?
@@ -708,6 +714,7 @@ module OneLogin
       # Validates if exists valid SubjectConfirmation (If the response was initialized with the :allowed_clock_drift option,
       # timimg validation are relaxed by the allowed_clock_drift value. If the response was initialized with the
       # :skip_subject_confirmation option, this validation is skipped)
+      # There is also an optional Recipient check
       # If fails, the error is added to the errors array
       # @return [Boolean] True if exists a valid SubjectConfirmation, otherwise False if soft=True
       # @raise [ValidationError] if soft == false and validation fails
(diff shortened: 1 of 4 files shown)


geneshuman:
I'm actually not certain if the Recipient & Destination fields should bother be the same.  Currently Destination is being matched against settings.assertion_consumer_service_url.  It may be the case that a separate setting should be created for matching the Recipient tag, although I'm not entirely sure.

pitbulk:
Can you add a flag tu let users disable  the Recipient check? (In order to maintain compatibility)

geneshuman:
OK, I've done this & added a bit of documentation.  I've set the skip flag to true by default for backwards compatibility.  Happy to do any refactoring/reworking of this that you'd prefer.

pitbulk:
I think the default state should be to not skip the recipient... and if users want to keep old behaviour... force the to use the skip_recipient_check

geneshuman:
OK, I've done this, however the change broke a handful of test.  The reason for this is that if you decode valid_response.xml.base64, you see that the Destination & Recipient messages aren't aligned.  And I don't think I can just change one of them & re-encode the message, as this modification will break the signatures in the message.  And I'm not really sure how to regenerate those signatures.  So for the tests in question, I skipped the recipient check.  I'm not sure how I feel about that, as these are the basic response validation tests in question.  But I'm not sure of an alternitive.  If somebody that knows more than me wants to rebuild valid_response.xml.base64, it would be an easy change to these tests afterwards.  Although changing that file may have some other un-intended consequences.

lib/onelogin/ruby-saml/response.rb
```diff
@@ -743,7 +738,7 @@ def validate_subject_confirmation
           next if (attrs.include? "InResponseTo" and attrs['InResponseTo'] != in_response_to) ||
                   (attrs.include? "NotOnOrAfter" and (parse_time(confirmation_data_node, "NotOnOrAfter") + allowed_clock_drift) <= now) ||
                   (attrs.include? "NotBefore" and parse_time(confirmation_data_node, "NotBefore") > (now + allowed_clock_drift)) ||
-                  (attrs.include? "Recipient" and !options[:skip_recipient_check] and attrs['Recipient'] != settings.assertion_consumer_service_url)
+                  (attrs.include? "Recipient" and !options[:skip_recipient_check] and attrs['Recipient'] != settings&.assertion_consumer_service_url)
```
pitbulk: It seems there is an extra & on that line

geneshuman:
No, that's intentional, its the new safe navigation operator
<http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/> for the case in
which settings is nil.

geneshuman:
Although maybe I shouldn't use it as it totally breaks your CI :P

pitbulk:
Thanks for contribute!

PR #400, opened 2017-05-26
validate_document_with_cert uses local cert when no cert in response

Adam21e:
I hit an issue when you have 2 certs signing and encryption and the cert is not included in the response.

After tracing it down, it appears that the validate_document_with_cert behaves differently to the validate_document method in that it won't fall back to the supplied cert when there was no cert included in the response.

I tried to write tests, but wasn't sure how to generate an signing/encryption pair of certs and wasn't sure I was allowed to use what my SP was using.

### DIFF
--- a/lib/xml_security.rb
+++ b/lib/xml_security.rb
@@ -260,9 +260,11 @@ module XMLSecurity
         # check saml response cert matches provided idp cert
         if idp_cert.to_pem != cert.to_pem
           return false
+        end
+      else
+        base64_cert = Base64.encode64(idp_cert.to_pem)
       end
-        validate_signature(base64_cert, true)
-      end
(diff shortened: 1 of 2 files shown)


pitbulk:
@Adam21e you can generate self-signed cert for the SP with that tool:
https://www.samltool.com/self_signed_certs.php

Adam21e:
Thanks, I had a lot more time today so poked around a bit more and found a similar test already and added a new test based on this.

response_test has a "**return true when no X509Certificate and the cert provided at settings matches**" which is the situation that I have.

The difference for me is that I am using **idp_cert_multi** and not **idp_cert**. 

Using **idp_cert_multi** triggers _validate_document_with_cert_ which is where the issue for me was.

PR #404, opened 2017-06-22
Add missing parenthesis to NotBefore error message

joshwetzel:
The `NotBefore` condition is missing a leading parenthesis in its error message.

On master:
> Current time is earlier than NotBefore condition TIME < TIME)

On branch:
> Current time is earlier than NotBefore condition (TIME < TIME)


pitbulk:
thanks!

pitbulk:
cryptosystem looks good, maybe in a future, we could use a 3rd party lib (or a fork of it) and delegate all the "magic"

joshwetzel:
Let me know what you had in mind. We rely pretty heavily on `ruby-saml` so I'd be happy to help out.

PR #418, opened 2017-09-20
Fix verification of SAML message signatures

jeffparsons:
## Status
?? **READY/IN DEVELOPMENT/HOLD**

## Description

Current implementation of `validate_signature` incorrectly assumes it can reconstruct the relevant URI parts (RelayState, SAMLRequest, SigAlg) from the _decoded_ values it receives. This contradicts the SAML specification, which requires computing the signature based on the URI components _as sent by the Identity Provider_.

The spec actually explicitly remarks that trying to reconstruct the URI parts is a bad idea because there is no single canonical URI encoding of a given value:

> Further, note that URL-encoding is not canonical; that is, there are multiple legal encodings for a given value. The relying party MUST therefore perform the verification step using the original URL-encoded values it received on the query string. It is not sufficient to re-encode the parameters after they have been processed by software because the resulting encoding may not match the signer's encoding.

I've preserved the old behaviour in the case that the caller doesn't provide the the correct (raw URI-encoded) parts, for backward compatibility.

This could probably use a couple of tests, but I figured I'd better see if you're ok with the general approach first.

## Todos

- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce

- Set up a Service Provider using ruby-saml, implementing IDP-initiated "Single Logout" with HTTP-Redirect binding.
- Register your app in Microsoft ADFS as a Relying Party.
- Try to log out from Microsoft's ADFS (IDP-initiated SLO).
- Observe that ruby-saml will erroneously consider the signature on the logout request to be incorrect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,35 @@
 # Ruby SAML [![Build Status](https://secure.travis-ci.org/onelogin/ruby-saml.svg)](http://travis-ci.org/onelogin/ruby-saml) [![Coverage Status](https://coveralls.io/repos/onelogin/ruby-saml/badge.svg?branch=master%0A)](https://coveralls.io/r/onelogin/ruby-saml?branch=master%0A) [![Gem Version](https://badge.fury.io/rb/ruby-saml.svg)](http://badge.fury.io/rb/ruby-saml)
 
+## Updating from 1.5.0 to 1.6.0
+
+Version `1.6.0` changes the preferred way to construct instances of `Logoutresponse` and `SloLogoutrequest`. Previously the _SAMLResponse_, _RelayState_, and _SigAlg_ parameters of these message types were provided via the constructor's `options[:get_params]` parameter. Unfortunately this can result in incompatibility with other SAML implementations; signatures are specified to be computed based on the _sender's_ URI-encoding of the message, which can differ from that of Ruby SAML. In particular, Ruby SAML's URI-encoding does not match that of Microsoft ADFS, so messages from ADFS can fail signature validation.
+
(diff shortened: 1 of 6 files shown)


jeffparsons:
Those test failures on some configs appear to have nothing to do with the code changes.

pitbulk:
Yes, it seems not related, we will need to apply the same at the [validate_signature](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/slo_logoutrequest.rb#L227) of the Logout Response.

As well as provide tests and documentation on the README.

jeffparsons:
@pitbulk I've written some tests, and made the equivalent change for `Logoutresponse`.

I'm not sure what kind of change you'd like to the README; I can't see any existing documentation about using `get_params`, so it's not as straightforward as just updating what's already there. Can you give me some pointers on where you'd like the documentation added, and how much detail you'd like?

pitbulk:
I think we may write an "Update from 1.5.0 to 1.6.0 version" and we ahould describe there that previously we  used in LogoutRequest and LogoutResponse constructor the optioms [get_params] but now we recommend to use raw_get_params... and maybe and example of code of the new use.

jeffparsons:
@pitbulk I've pushed some documentation in `README.md`. Does that look ok to you?

jeffparsons:
Cheers. 😎

jeffparsons:
Is there anything I can do to help move this (and https://github.com/onelogin/ruby-saml/pull/420) closer to landing?

jeffparsons:
No drama whatsoever. 😃  Cheers!

PR #425, opened 2017-11-30
Allow to set "AllowCreate" on name identifier.

jodigiordano:
As a first step on implementing SAML in our codebase, we do not want to create new users automatically.
So we had a need to be able to tell the Identity Provider we won't permit creating new users.

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -111,20 +111,26 @@ module OneLogin
         if settings.assertion_consumer_service_url != nil
           root.attributes["AssertionConsumerServiceURL"] = settings.assertion_consumer_service_url
         end
+
         if settings.issuer != nil
           issuer = root.add_element "saml:Issuer"
           issuer.text = settings.issuer
         end
+
         if settings.name_identifier_format != nil
-          root.add_element "samlp:NameIDPolicy", {
-              # Might want to make AllowCreate a setting?
-              "AllowCreate" => "true",
-              "Format" => settings.name_identifier_format
+          atributes = {
+            "Format" => settings.name_identifier_format,
+            "AllowCreate" => "true"
           }
+
+          if settings.name_identifier_allow_create != nil
+            atributes["AllowCreate"] = settings.name_identifier_allow_create.to_s
+          end
+
+          root.add_element "samlp:NameIDPolicy", atributes
         end
 
         if settings.authn_context || settings.authn_context_decl_ref
-
           if settings.authn_context_comparison != nil
             comparison = settings.authn_context_comparison
           else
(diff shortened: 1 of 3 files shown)


pitbulk:
The "AllowCreate" parameter isn't used by SP to indicate IdP it won't permit creating new users.

> AllowCreate
 [Optional]
A Boolean value used to indicate whether [E14]the requester grants to the identity provider, in the 
course of fulfilling the request, permission to create a new identifier or to associate an existing 
identifier representing the principal with the relying party. Defaults to "false"
if not present or the entire 
element is omitted.
The 
AllowCreate
 attribute may be used by some deployments to influence the creation of state 
maintained by the identity provider pertaining to the use of a name identifier (or any other persistent, 
uniquely identifying attributes) by a particular relying party, for purposes such as dynamic identifier or 
attribute creation, tracking of consent, subsequent use of the Name Identifier Management protocol 
(see Section 
3.6
), or other related purposes.
When “false”, the requester tries to constrain the identity provider to issue an assertion only if such 
state has already been established or is not deemed applicable by the identity provider to the use of 
an identifier. Thus, this does not prevent the identity provider from assuming such information exists 
outside the context of this specific request (for example, establishing it in advance for a large number 
of principals).
A value of “true” permits the identity provider to take any related actions it wishes to fulfill the request, 
subject to any other constraints imposed by the request and policy (the 
IsPassive
 attribute, for 
example).
Generally, requesters cannot assume specific behavior from identity providers regarding the initial 
creation or association of identifiers on their behalf, as these are details left to implementations or 
deployments. Absent specific profiles governing the use of this attribute, it might be used as a hint to 
identity providers about the requester’s intention to store the identifier or link it to a local value.
A value of “false” might be used to indicate that the requester is not prepared or able to do so and 
save the identity provider wasted effort. Requesters that do not make specific use of this attribute SHOULD generally set it to “true” to maximize interoperability"

We set it that value to true always to maximize interoperability.

Reference Pag 53
https://www.oasis-open.org/committees/download.php/56776/sstc-saml-core-errata-2.0-wd-07.pdf

pitbulk:
In addition, that toolkit is used to process and validate a SAMLResponse and extract NameID and User attributes.  What you do with that info is our of SAML control, you decide in your app what do you want to do with that info.

jodigiordano:
> The "AllowCreate" parameter isn't used by SP to indicate IdP it won't permit creating new users.

Oh I see, I misunderstood the purpose of this parameter. It's the other way around: it tells the IdP do not create a new identifier on its side.

> What you do with that info is our of SAML control, you decide in your app what do you want to do with that info.

Yes, that I understand.
It was kind of a courtesy to the IdP to let it know of our intent.

> We set it that value to true always to maximize interoperability.

I do understand why but I'm not convinced it is the right thing to do.
Someone could read the specification (better than me) and could want to tell the IdP to not create new identifiers (even if the IdP can ignore the request).
Right now the library makes a decision on this parameter but do not tells why.
Maybe that interoperability comment should be in the code instead of `# Might want to make AllowCreate a setting?`.

Please close this issue if you think `AllowCreate` should always be `true`.

Cheers!

pitbulk:
The main goal of the toolkit is to keep the settings as simple as possible. Adding that new settings that dont really help the SP developer makes no much sense.

PR #434, opened 2018-01-26
Allow formatting of certificates that contain \r

id4ho:
If an idp_cert contains a '\r' it can blow up upon response validation
with `OpenSSL::X509::CertificateError: nested asn1 error` even if the
cert is otherwise valid (or would have been post-formatting).

From the way `OneLogin::RubySaml::Utils.format_cert` is implemented it
would appear that it *is* expected for '\r's to be present since it
tries to strip them appropriately during the formatting below the guard
statement. Unfortunately, the guard statement at the top short circuits
the formatter when certificates contain '\r' since:
```
irb:0> "asldfkj\r".match(/\x0d/)
=> #<MatchData "\r">
```

Removing the `cert.match(/\x0d/)` doesn't actually break any specs but
from the comment it seems that it may have been intended to ensure
that encoded certs (i.e. .der) are not run through the formatter. I've
added a `.der` cert to `tests/certificates` and asserted that it isn't
changed when run through `format_cert` by checking for `ascii_only?`.


lib/onelogin/ruby-saml/utils.rb
```diff
@@ -22,7 +22,7 @@ class Utils
       #
       def self.format_cert(cert)
         # don't try to format an encoded certificate or if is empty or nil
-        return cert if cert.nil? || cert.empty? || cert.match(/\x0d/)
+        return cert if cert.nil? || cert.empty? || !cert.ascii_only?
```
jackvnimble: Maybe I've misinterpreted the comment above and this function isn't meant to support .der in which case.. maybe we can simply remove `|| cert.match(/\x0d/)`?

PR #438, opened 2018-02-06
Don't output empty URL parameters

### LINKED ISSUE #437: Creating AuthRequests with nil RelayState sends empty RelayState URL param
Doing `OneLogin::RubySaml::Authrequest.new.create(settings, RelayState: nil)` still creates an empty RelayState url param. Ideally nil (or perhaps even falsy) RelayState would cause `ruby-saml` to not output a param. This caused an issue for a customer integration recently. SimpleSaml (the PHP library) apparently decided that we had an improperly signed request with the empty param.

vincentwoo:
Fixes #437

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -75,7 +75,7 @@ module OneLogin
           params['Signature'] = encode(signature)
         end
 
-        params.each_pair do |key, value|
+        params.compact.each_pair do |key, value|
           request_params[key] = value.to_s
         end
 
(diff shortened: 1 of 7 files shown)


vincentwoo:
Er, this depends on ruby 2.4.0+. I assume that is not acceptable?

vincentwoo:
Okay, I used the old way of doing compact.

pitbulk:
that will affect not only RelayState but other parameters.

what about after
```
params.each_pair do |key, value|
    request_params[key] = value.to_s
end
```

check if request_params['RelayState'] exists and if empty and if that is the case, remove it

vincentwoo:
I don't think the logic should be special cased to RelayState. I think any empty param can trigger the signature validation issue on not-quite-great SAML libraries, and it's hard to imagine the utility of any empty param in any of these auth requests.

vincentwoo:
ping

pitbulk:
Sorry for the delay.

What I'm worried is with the fact that maybe devs are sending params to the create method not related with SAML, and that PR that change the behavior can break their environments.

Issue #440, opened 2018-02-13
Invalid audience: Misleading error message


capybaras:
If an SP configuration expects an audience of "Expected", but the SAML assertion contains an audience of "Actual" you will get the following error message:

> "Expected is not a valid audience for this Response - Valid audiences: Actual"

https://github.com/onelogin/ruby-saml/blob/d500e8f2d65da89b2fe66c3e253e783b540f8f77/lib/onelogin/ruby-saml/response.rb#L600

The message should be flipped.

> "Invalid Audience. The audience 'Actual', does not match the expected audience of 'Expected'"

akshah123:
Part of the confusion here is also that `issuer` is used to verify `audience`.

If this project is used from purely an SP perspective, using `issuer` to check against `audience` seems a little strange.

@capybaras do you have any insight into why that is?

pitbulk:
Im agree that message is less confuse. But worried about breaking environments with that change. I will merge that in a new release properly versioned.

pitbulk:
@capybaras In SAML the "issuer" is the entity that generated the SAML message.

Im agree with you on the issuer/audience stuff... but was implemented like that at the begining of the project.

nbaki:
Yes, looks like issuer is mapping to audience on the IDP end, this shouldn't be the case.

capybaras:
The mixing of issuer and audience is a different issue, so perhaps we could open a separate issue for that one.

@pitbulk waiting for a new version makes sense

capybaras:
@pitbulk now that the 1.7.x security fixes are in, what version do you think this belongs in?

PR #441, opened 2018-02-13
Updated invalid audience error, and tests to match.

capybaras:
## Fixes:
https://github.com/onelogin/ruby-saml/issues/440

## Example
Given a SAML Assertion with the Audience value of "A", and an SP configuration requiring an audience of "B"

### Old message:

> "B is not a valid audience for this Response - Valid audiences: A"

or (for multiple)
> "B is not a valid audience for this Response - Valid audiences: A1, A2, A3"

### New message:

> Invalid Audience. The audience A, did not match the expected audience B

or (for multiple)

> Invalided Audiences. The audiences A1, A2, A3 did not match the expected audience B


mokhan:
I like the new error messaging better. The old error message confused me on several different occasions.

PR #446, opened 2018-02-28
OneLogin::RubySaml::Utils.element_text should normalize text before returning it

### LINKED ISSUE #445: Fix for CVE-2017-11428 does not normalize text before returning it
[The fix for CVE-2017-11428](https://github.com/onelogin/ruby-saml/commit/048a544730930f86e46804387a6b6fad50d8176f) essentially replaced the call to `element.text` with [`element.texts.join`](https://github.com/onelogin/ruby-saml/commit/048a544730930f86e46804387a6b6fad50d8176f#diff-d3e2c1bdc6c1e8b207dd7ded3ae35320R289). However, the behavior of `text` differs from `texts` in the following way (**emphasis** mine):
> This method returns the value of the first text child node, which ignores the `raw` setting, so **always returns normalized text**.
> https://ruby-doc.org/stdlib-2.3.0/libdoc/rexml/rdoc/REXML/Element.html#method-i-text

`texts` does not normalize the text before returning it. This primarily breaks cases where the raw text contains elements that are escaped. Here's an example of the main difference:

```ruby
REXML::Document.new('<foo>the pen &gt; the sword</foo>').elements.first.text
# => "the pen > the sword"
REXML::Document.new('<foo>the pen &gt; the sword</foo>').elements.first.texts.join
# => "the pen &gt; the sword"
```

### Proposed fix

Since [`REXML::Element#text` calls `REXML::Text#value`](https://github.com/ruby/ruby/blob/9af2ab9d89acb693236e9da58a482551e640e780/lib/rexml/element.rb#L453) to normalize the text value, change the call to `elements.texts.join` with `elements.texts.map(&:value).join` to normalize the text before returning it:

```ruby
REXML::Document.new('<foo>the pen &gt; the sword</foo>').elements.first.texts.map(&:value).join
# => "the pen > the sword"
```

I can look into creating a pull request for this fix if you agree.

Umofomia:
Fixes #445.

### DIFF
--- a/lib/onelogin/ruby-saml/utils.rb
+++ b/lib/onelogin/ruby-saml/utils.rb
@@ -286,7 +286,7 @@ module OneLogin
       # that there all children other than text nodes can be ignored (e.g. comments). If nil is
       # passed, nil will be returned.
       def self.element_text(element)
-        element.texts.join if element
+        element.texts.map(&:value).join if element
       end
     end
   end
--- a/test/utils_test.rb
+++ b/test/utils_test.rb
@@ -230,6 +230,16 @@ class UtilsTest < Minitest::Test
         assert_equal 'element & text', OneLogin::RubySaml::Utils.element_text(element)
       end
 
+      it 'returns the CDATA element text' do
+        element = REXML::Document.new('<element><![CDATA[element & text]]></element>').elements.first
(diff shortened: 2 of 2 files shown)


Umofomia:
It looks like the Travis CI build is failing for jruby, but it appears to be failing for reasons unrelated to my change. The same tests appear to be failing in `master` anyway.

PR #452, opened 2018-03-16
Fix behavior of skip_conditions flag on Response

### LINKED ISSUE #451: Saml Response validation fails unexpectedly if there is no samlp:Response/saml:Assertion/saml:Conditions element and the skip_conditions option is true
When there is no `samlp:Response/saml:Assertion/saml:Conditions` element in the Saml Response, validation will still fail if skip_conditions is true, because `OneLogin::RubySaml::Response#validate_one_conditions` doesn't respect the `skip_conditions` flag like `OneLogin::RubySaml::Response#validate_conditions` does.

I'm happy to submit a pull request to fix it if you guys are open to that. It looks like an appropriate test xml file is already in the code base.

Thanks,

Matt

soupmatt:
Ensure the skip_conditions flag allows a Response with no Conditions
elements to be valid. Fixes #451

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -615,10 +615,13 @@ module OneLogin
       end
 
       # Checks that the samlp:Response/saml:Assertion/saml:Conditions element exists and is unique.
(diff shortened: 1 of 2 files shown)


soupmatt:
Thanks for the quick response. 🍻

PR #453, opened 2018-03-23
Raise a more descriptive exception if idp_sso_target_url is missing

tomilaine:
If idp_sso_target_url is missing, creating the authrequest fails currently with:
```NoMethodError (undefined method `+' for nil:NilClass)```
This is not very descriptive, so this pull request makes the exception message more helpful.

The motivation/need for this is that typically services handle the saml settings in some kind of dynamic way. If something goes wrong there and the settings are missing, it's good that ruby-saml will point them to the right direction.


oTsogbadrakhChinzorig:
It’s happened to me, but still saml 2.0 requires sso url. We can change messages but from sp side it doesn’t have any meaning

PR #454, opened 2018-03-27
Added Response available options list

mberlanda:
I recently did a PR in `omniauth-saml` gem to allow more options for a `OneLogin::RubySaml::Response` object (https://github.com/omniauth/omniauth-saml/pull/159).

After merging it, I was discussing with @md5 about the possibility to source the allowed options from `ruby-saml` gem instead of hardcode them in this gem. 

As of today, [`omniauth-saml`](https://rubygems.org/gems/omniauth-saml/) had ~2.3 million downloads and I believe that introducing a whitelist of response options will make the dependency on `ruby-saml` stronger and even more reliable.

Moreover, it will allow to developers who are using `ruby-saml` standalone to have a better overview on the options they can use.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -30,6 +30,15 @@ module OneLogin
 
       attr_accessor :soft
 
+      # Response available options
+      # This is not a whitelist to allow people extending OneLogin::RubySaml:Response
+      # and pass custom options
+      AVAILABLE_OPTIONS = [
+        :allowed_clock_drift, :check_duplicated_attributes, :matches_request_id, :settings, :skip_authnstatement, :skip_conditions,
+        :skip_destination, :skip_recipient_check, :skip_subject_confirmation
+      ]
+      # TODO: Update the comment on initialize to describe every option
+
       # Constructs the SAML Response. A Response Object that is an extension of the SamlMessage class.
       # @param response [String] A UUEncoded SAML response from the IdP.
       # @param options  [Hash]   :settings to provide the OneLogin::RubySaml::Settings object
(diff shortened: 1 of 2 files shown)


mberlanda:
hi @pitbulk 

I saw in the travis build that `jruby-1.7.21` and `jruby-9.0.0.0` are failing since a while. They might need a separated Gemfile (https://docs.travis-ci.com/user/languages/ruby/#JRuby:-C-extensions-are-not-supported).

In `ree` and `ruby-1.8.7` an array of symbols is raising `TypeError: Symbol as array index`. How are you handling this data structure?

Thanks

pitbulk:
My unique concern about merging it is:
- What happens to projects using ruby-saml that extended the SAMLResponse options with more option? This PR will break their projects.

Related to jruby, I haven't got an environment to test it.

mberlanda:
Thank you for your quick reply!

Actually the whitelist of options is a nice workaround to make sure that dependent gems such as `omniauth-saml` will be updated with the options offered by your gem. I haven't considered the case of someone extending the class and passing custom options.

This is not the only possible solution but it would prevent several prs on other gems which depends on yours.

Would it be suitable to you to rename the `ALLOWED_OPTIONS` to `AVAILABLE_OPTIONS` and remove the validation of the options provided? This would achieve the purpose of this PR.

pitbulk:
Yes, I think adding ALLOWED_OPTIONS is not bad idea

mberlanda:
Perfect! I changed then to `AVAILABLE_OPTIONS` and dropped the commit which was introducing the filter. I guess this could be ready for a review.

mberlanda:
hi @pitbulk ! Could you please bump a minor version of the gem including this PR changes?
Thank you very much

PR #458, opened 2018-05-29
Remove ruby 2.4+ warnings

davidlibrera:
I do some code refactor in order to remove warnings about unused variables, inner scoped variables with same name of outer scope and a missing initialization of ```@logger``` instance attribute inside Logger. I don't change any code behaviour.


lib/onelogin/ruby-saml/response.rb
```diff
@@ -288,7 +289,7 @@ def issuers
             raise ValidationError.new(error_msg)
           end
 
-          doc = decrypted_document.nil? ? document : decrypted_document
+          _doc = decrypted_document.nil? ? document : decrypted_document
```
pitbulk: Can you remove this line? It seems not required.

lib/xml_security.rb
```diff
@@ -318,10 +318,10 @@ def validate_signature(base64_cert, soft = true)
 
       # check digests
       ref = REXML::XPath.first(sig_element, "//ds:Reference", {"ds"=>DSIG})
-      uri = ref.attributes.get_attribute("URI").value
+      _uri = ref.attributes.get_attribute("URI").value
```
pitbulk: Can you remove this line as well?

davidlibrera:
I commit the fix. Should I squash it into one commit?

pitbulk:
I think we are ok

PR #460, opened 2018-05-29
Support parsing metadata of multiple IdPs at once

tkalliom:
I wanted to discover all IdPs involved in a federation, but `IdpMetadataParser` only supported getting one at a time. This introduces new methods which return all found IdPs as an array, not changing established interface.

Working on this I also stumbled on the bug that ”implicitly choose first found IdP” failed if there were SP `EntityDescriptor`s before IdP ones. The refactoring commit below the new functionality fixes this and adds a corresponding test case.

### DIFF
--- a/lib/onelogin/ruby-saml/idp_metadata_parser.rb
+++ b/lib/onelogin/ruby-saml/idp_metadata_parser.rb
@@ -145,7 +145,7 @@ module OneLogin
         end
 
         get = Net::HTTP::Get.new(uri.request_uri)
-        response = http.request(get)
+        @response = http.request(get)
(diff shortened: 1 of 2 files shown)


pitbulk:
I can't merge this PR since it can break integrations using, for example, the get_idp_metadata method.

It's ok if you refactor code, but you should maintain the public methods of the existing class.

tkalliom:
I don’t quite follow, why would using the constants no longer be possible?

```
> OneLogin::RubySaml::IdpMetadataParser::METADATA
=> "urn:oasis:names:tc:SAML:2.0:metadata"
```

PR #465, opened 2018-07-03
Partial attributes to security settings doesn't nullify missing ones

davidlibrera:
I supposed this was a bug, because if I want to enable metadata signing I need to provide all attributes, while I just want to enable it without change other attributes.

### DIFF
--- a/lib/onelogin/ruby-saml/settings.rb
+++ b/lib/onelogin/ruby-saml/settings.rb
@@ -10,7 +10,9 @@ module OneLogin
     #
     class Settings
       def initialize(overrides = {})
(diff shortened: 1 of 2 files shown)


pitbulk:
I think you are right and your approach is the expected behavior.
The problem right now with that PR is that can break current integrations.

In order to maintain compatibility, what do you think if we add a second parameter, by default the old behavior, to control when or not respect security values?

davidlibrera:
Great! I changed the method with https://github.com/onelogin/ruby-saml/pull/465/commits/b37366e1b7a3641e8aac0ba868395a6de3b11e7f and I added another example in order to test both behaviours.

Do you agree?

PR #468, opened 2018-07-17
same code as authnrequest for logoutrequest

davidlibrera:
Having separated method for document generation and document signing (like as actual authnrequest implementation) allow a simply way to override document structure before signing it.


pitbulk:
Can't merge it since this breaks integrations that expects that the create_logout_request_xml_doc method returns signed XMLs

davidlibrera:
No problem but I'll need a few days

davidlibrera:
@pitbulk Do you have news about this PR?

PR #471, opened 2018-09-03
Allow for `allowed_clock_drift` to be set as a string

stavros-wb:
Hello team.

SAML settings maybe configured as strings instead of having types on them. I thought we should cast `allowed_clock_drift` to a float before using it. A use case might users keeping their SAML settings in string only stores like redis or pg hstore


stavros-wb:
Hello @pitbulk, can you please tag this? we want to deploy the change and are a little blocked because this is a gemspec dependency for us. Thank you in advance

pitbulk:
Done, new 1.9.0 version available.

PR #475, opened 2018-10-11
Update README with changes related to v. 1.9.0

MrBerg:
I assume 1.9.0 is the newest stable, so it needs an updated README in my opinion (and maybe a tag?). I went through the commits made after 1.8.0 up until the 1.9.0 version release, as well as the changelog, and wrote down the changes I saw as important.


pitbulk:
Ty

PR #493, opened 2019-03-21
Explicit about the error when signed element is not found and found m…

perryism:
# Purpose
It is unclear what goes wrong from a caller's stand point when it only gets the generic message "Invalid Signature on SAML Response".   

This PR is to set the error message about what goes wrong when the signed element id is found more than once or not found.


mathildemouw:
This is great! This would help us figure out when something's going wrong with our customers' SAMLResponses much more easily 👍

PR #495, opened 2019-03-27
Add ruby 2.6 to CI

bwillis:
We were investigating a Ruby 2.6 upgrade in our repo and noticed this didn't yet support it, so I thought we would start a PR to see if there are any issues that come up. In our repo, we did see some specs fail around unsigned request that I hope will get surfaced here.

PR #497, opened 2019-04-05
Make CI 💚

bwillis:
In creating #495 I noticed there was a lot of repeat failures in the CI pipeline. They seemed to stem from 2 issues:
 - Running Ruby versions against an old unsupported version of Nokogiri
 - Failing specs due to ruby-saml not being compatible with Ruby v1.8.7 anymore

This PR tries to resolve both those issues by excluding incompatible ruby/nokogiri builds and removing support for ruby 1.8.7 in order to make CI 💚

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: ruby
 rvm:
-  - 1.8.7
   - 1.9.3
   - 2.0.0
   - 2.1.5
@@ -9,7 +8,6 @@ rvm:
   - 2.3.0
   - 2.4.0
   - 2.5.0
-  - ree
   - jruby-1.7.27
   - jruby-9.1.17.0
   - jruby-9.2.0.0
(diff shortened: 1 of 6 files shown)


bwillis:
@pitbulk what do you think?

pitbulk:
Rather than remove compatibility with 1.8.7, this compatibility issue should be fixed.

pitbulk:
Now it pass travis
https://github.com/onelogin/ruby-saml/commit/71cb1e23c8ce3b93b47ec2049e66ac06426fc008

bwillis:
Looks good, for reference the fix for Ruby 1.8.7 support was in 9e6849e0ef10bb303851f15adf2e049cfe5fd94a.

> I know 1.8.7 is kinda deprecated, but there are a lot of people still using it so I may keep its support.

I mean is it even "kinda" deprecated 😀? It's been EOL since [before 2015](https://www.ruby-lang.org/en/downloads/branches/).

Anyway, your call. Thanks for fixing!

PR #498, opened 2019-04-11
Implementing Process Transform. Related with 

### LINKED ISSUE #496: digest doesn't match unless C14N EXCLUSIVE is used for canonicalization
I'm trying to validate a  `SAMLRequest`, but   I get the error "digest mismatch".

In troubleshooting I notice that if I hack `ruby-saml` to use `Nokogiri::XML::XML_C14N_EXCLUSIVE_1_0` just for the digest canon method then the `SMLRequest` DOES validate:

```
      # check digests
      ref = REXML::XPath.first(sig_element, "//ds:Reference", {"ds"=>DSIG})

      hashed_element = document.at_xpath("//*[@ID=$id]", nil, { 'id' => extract_signed_element_id })
      
     # HACK:  force EXCLUSIVE to troubleshoot
      canon_algorithm = Nokogiri::XML::XML_C14N_EXCLUSIVE_1_0
      canon_hashed_element = hashed_element.canonicalize(canon_algorithm, inclusive_namespaces)
```

In the response the canonicalization method is specified as http://www.w3.org/TR/2001/REC-xml-c14n-20010315, which is used for the signature and and works:

```
  <Signature xmlns="http://www.w3.org/2000/09/xmldsig#">
    <SignedInfo>
      <CanonicalizationMethod Algorithm="http://www.w3.org/TR/2001/REC-xml-c14n-20010315" />
```

I suspect there is a bug in `ruby-saml` b/c the same `SAMLResponse` validates using the online saml validation tool: https://www.samltool.com/validate_response.php

What could cause it to validate using the online tool, but fail with digest_match in ruby-saml ?

pitbulk:
Should fix #496 496

### DIFF
--- a/lib/xml_security.rb
+++ b/lib/xml_security.rb
@@ -326,6 +326,9 @@ module XMLSecurity
         '//ds:CanonicalizationMethod',
         { "ds" => DSIG }
       )
+
+      canon_algorithm = process_transforms(ref, canon_algorithm)
+
       canon_hashed_element = hashed_element.canonicalize(canon_algorithm, inclusive_namespaces)
 
       digest_algorithm = algorithm(REXML::XPath.first(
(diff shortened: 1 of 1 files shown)


robinator:
@pitbulk Is this waiting on examples for testing or something else?

pitbulk:
Yes, I was expecting to get some examples to add to the tests. I will try to get them actively.

PR #519, opened 2019-11-06
Update single logout code sample to encourage early session termination

jsdalton:
## Background

Recently, our team encountered a security issue reported by one of our customers. The customer complained that users were not being logged out properly of our application (a service provider) when using single logout.

An investigation revealed a proximate cause: this customer's IdP incorrectly implemented the single logout specification, and did not redirect the user back to us (the service provider) after successful completion of the single logout process. The specification indicates the IdP _must_ do this; nevertheless the IdP was not in compliance.

The problem this surfaced on our side is that our procedure for terminating the session is quite fragile. Because we terminated the session on successfully receiving the `LogoutResponse` from the IdP, this means that if anything goes wrong -- from the time the initial `LogoutRequest` is created on our side, to the process the IdP takes to log the user out, to the final response sent back to us -- the end result is the user is not logged out of our service, in spite of initiating a log out (i.e. clicking the logout button on our website).

We reviewed the SAML 2.0 spec this morning once again to see if it provided guidance on this. Though it does provide very clear guidance about the rules for a service provider (i.e. "Session Participant") when receiving a `LogoutRequest` (session participant must terminate the session), it is totally silent about when the session should be terminated for logouts initiated by the session participant, i.e. whether on initial creation of the logout request or on subsequent receipt of a logout response from the IdP.

The closest language we could find was in section 3.7.1, the first sentence:

> A session participant or session authority sends a `<LogoutRequest>` message to indicate that a session has been terminated.

This at least hints at the idea that the session participant first terminates a session before sending the `LogoutRequest`, albeit not very explicitly.

## PR Change

On further reviewing our own code, we realized that the main reason we had not been terminating the session before generating the `LogoutRequest` was in fact that we had copied the code in your README example as a starting framework. Although it has evolved and had tests added, fundamentally we follow the logic presented here.

Given the above, we feel that the proposed update would benefit future developers. Since many developers in the Ruby world rely on this gem to implement the basics of SAML-based SSO and SLO, it makes sense to propose in example code a solution that is reliable, secure, and in adherence with the specification.

The proposed change to the example code loosely reflects the change we ended up making to our own production code. This not only solved the problem for our customer with the problematic implementation described above, but also helped solve the other problem mentioned where customers who encountered arbitrary errors were not successfully logged out from our service.

If there are use cases or problems with the suggestion here, or if you have reason to believe this proposal is not in accordance with the SAML spec, I would have course be happy to hear the feedback so that wec can ensure our own updates meet the standards.

Thanks for considering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,15 +616,20 @@ def sp_logout_request
     delete_session
   else
 
-    # Since we created a new SAML request, save the transaction_id
-    # to compare it with the response we get back
     logout_request = OneLogin::RubySaml::Logoutrequest.new()
-    session[:transaction_id] = logout_request.uuid
-    logger.info "New SP SLO for userid '#{session[:userid]}' transactionid '#{session[:transaction_id]}'"
 
     if settings.name_identifier_value.nil?
       settings.name_identifier_value = session[:userid]
     end
+
+    # Ensure user is logged out before redirect to IdP, in case anything
+    # goes wrong during single logout process
+    delete_session
+
+    # Since we created a new SAML request, save the transaction_id
+    # to compare it with the response we get back
+    session[:transaction_id] = logout_request.uuid
+    logger.info "New SP SLO for userid '#{session[:userid]}' transactionid '#{session[:transaction_id]}'"
 
     relayState =  url_for controller: 'saml', action: 'index'
     redirect_to(logout_request.create(settings, :RelayState => relayState))
@@ -654,6 +659,9 @@ def process_logout_response
   else
     # Actually log out this session
     logger.info "Delete session for '#{session[:userid]}'"
+
+    # Session should already have been deleted when generating SLO
+    # LogoutRequest; this is here as a failsafe
     delete_session
   end
 end


pitbulk:
The "SLO" feature has a big controversial.
You only need to read some of those threads:
- https://wiki.shibboleth.net/confluence/display/CONCEPT/SLOIssues
- https://www.portalguard.com/blog/2016/06/20/saml-single-logout-need-to-know/

I took the approach of not closing the local session before initiating the SLO process in order to let the user know, in case that the SLO flow breaks, that something happened, since the session on the ap will be active.   Otherwise, the user can believe that the SLO worked, but in fact, only the local session was closed.    

Imagine a scenario where the SLO process is broken and the IdP session keep alive, if you close local session for the user , user will believe all is good, and with a simple click to log in the user again, it will SSO (due active session on IdP). For him gonna be kinda unexpected.

jsdalton:
@pitbulk Thanks for the feedback.

I 100% agree with your thoughts on SLO. We actively discourage our customers from using it, but still offer support for those who request it. It's a bummer that the SAML spec is so poor in this regard. In fact, our team scoured the SAML 2.0 spec for an answer to the question of when the session should be terminated by the initiating service provider, and it's absolutely silent on this question -- in spite of spelling out in detail the responsibilities of the IdP as well as other SPs.

Anyhow, in real life we encountered a very serious flaw with the flow suggested by the example code. Somewhat frequently, a user would encounter an error during the SLO process. Really it could happen at any step -- especially since the spec envisions a number of redirects across many services. Regardless of where the error happened, the fundamental problem was this: **The user could not log out of our application.** Because the only way a successful logout could occur was if the user made it successfully _back_ to our application. Any flaw resulted in probably the worst outcome possible.

This was our primary driver for reworking our SLO process to ensure session termination occurs before anything. If an error occurs later, a user usually _sees_ evidence of that error. And in practice (not sure how others do this) we have a "logout" button which 99% of users just assumes logs a user out of the application, irrespective of other side effects. I don't think any of our customers or users would prefer that the user was not logged out in the event of an error, silent or not.

pitbulk:
There is an easy solution to not block the user logout process.
Before initiating the process of SLO, save a flag on the user session to indicate that the SLO process was initiated, and then execute the LogoutRequest.

Later you can add a check on that logout endpoint to verify that there is no SLO flag  enabled before executing this process, Otherwise, redirect the user to an intermediate view that notifies the user that a previous SLO flow was initiated and failed so the app now will execute a local logout (instead a global logout).

That way you notify properly the user as well as offer a way to close local session even when SLO process failed.

What do you think about this approach?

jsdalton:
Thanks again for the response @pitbulk 

To be clear about one thing, we are at this stage 100% satisfied with our current implementation. It passed both our internal security review as well as an external security review from a customer (large financial institution with strict security validation measures).

My motivation in opening this PR is to perhaps persuade your team that the example implementation here is not secure -- or rather to be more precise, leads to an outcome which potentially exposes end users to a not secure situation. Because this is a widely used library in the Ruby community I think it's rather important that the examples do promote secure solution, since I'm sure they are used as a template by many developers.

Having said the above, let me go back to your original reply:

> I took the approach of not closing the local session before initiating the SLO process in order to let the user know, in case that the SLO flow breaks, that something happened, since the session on the app will be active. Otherwise, the user can believe that the SLO worked, but in fact, only the local session was closed.
> 
> Imagine a scenario where the SLO process is broken and the IdP session keep alive, if you close local session for the user , user will believe all is good, and with a simple click to log in the user again, it will SSO (due active session on IdP). For him gonna be kinda unexpected.

I think you present here a valid use case for where an SLO process fails at some step, and the user expect it to have succeeded, and doesn't realize they are only logged out of the SP application but not the IdP (and thus could trivially be logged back in)

What if we considered approach where we _do_ handle the use case above, but at the same time continue to ensure that the user is logged out of the SP before initiating the SLO redirects. So more or less exactly what you suggest in your most recent reply:

1. User clicks log out.
2. User session is terminated (very important! 😄 )
3. Save "SLO initiated" flag to session as you suggest. Note for most frameworks (at least Rack-based like Sinatra or Rails) this flag would be saved on a NEW (anonymous session).
4. Redirect to IdP, complete SLO, etc.
5. On successful completion, back at SP with a SingleLogoutResponse, display whatever "everything worked" message as normal. If error but SP still successfully receives SingleLogoutResponse, display error message as normal. In either case, clear the "SPO initiated" flag (or mark as completed or whatever).
6. On unsuccessful SLO (e.g. something got disrupted, user was not logged out of IdP, never made it back to the app), if the user returns later to the SP application, SP can check the session for the "SLO initiatied" flag and display an error message, notification, what have you.

So really, I guess I'm recommending the same thing as you suggest, but doing so in the context of an anonymous session with a saved value vs. the active session which the user intended to terminate.

pitbulk:
Ok, Do you want to improve the PR with the solution we detailed here or do you want me to merge this PR?

jsdalton:
🤔 I think it would be slightly beyond the scope of this PR to handle item #6. Because this would likely have to happen at a different part of the applicaton, e.g. I try to visit page X, and the app sees that I'm logged out, but then the app _also_ checks for the presence of this flag so that it can display the error message along with (or instead of?) the login screen.

In general this PR already forms the basis that this solution builds on, because it sets the session value _after_ deleting the authenticated session, so there already exists a value that can be checked if the user visits a different page outside the SLO log out flow.

I would recommend to merge the PR as stands (barring any other feedback) since, per reasons above, we believe ensuring session termination is of primary importance.

pitbulk:
[saml2int](https://kantarainitiative.github.io/SAMLprofiles/saml2int.html#_single_logout_2) seems to recommend this behavior at
[SDP-SP34] so added the early session termination

PR #525, opened 2020-01-21
Relax nokogiri for Ruby 2.1

johnnyshields:
Recent change on master broke my bundle


johnnyshields:
Please merge. master is in a broken state now for Ruby 2.1/2.2

pitbulk:
There are security issues with old version of nokogiri, that why when possible, we forcing for the latest compatible when possible.

The typo will be fixed at: https://github.com/onelogin/ruby-saml/pull/550

johnnyshields:
@pitbulk that is an anti-pattern. The Ruby SAML gem should not "force" others to upgrade Nokogiri; users should do that on their own at their own discretion.

PR #536, opened 2020-04-17
Fetch attribute value

JDrizzy:
After setting up an AD-FS SSO and trying to pass a users first and last name through attributes, I found that AD-FS passes the following attribute names which are annoying/ugly to match:
```Ruby
# Attributes
{
  'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname' => ['Jarrad'],
  'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname' => ['M'],
}
saml_response.attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname']
# Jarrad
```

I _thought_ a `#fetch` might be ideal here to handle cases where one might want to retrieve an attribute value using regular expression?
```Ruby
# Attributes
{
  'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname' => ['Jarrad'],
  'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname' => ['M'],
}
saml_response.attributes.fetch(/givenname/)
# Jarrad
saml_response.attributes.fetch(/surname/)
# M
```

Happy to hear your thoughts ✌️


pitbulk:
I believe that makes sense. Can you update the documentation of the Readme as well?

PR #538, opened 2020-05-19
Reduce size (in bytes) of built gem

jaredbeck:
Omit documentation and tests from the gem build.
Reduces build size by ~1.5 MB.

### DIFF
--- a/ruby-saml.gemspec
+++ b/ruby-saml.gemspec
@@ -15,7 +15,9 @@ Gem::Specification.new do |s|
     "LICENSE",
     "README.md"
   ]
-  s.files = `git ls-files`.split("\n")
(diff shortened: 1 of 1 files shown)


jaredbeck:
>  Reduces build size by ~1.5 MB.

To be really (unnecessarily) specific, it's the install size, not the build size that is reduced. (the contents of the `.gem` file are decompressed during installation)

pitbulk:
I'm ok not including the tests in the gem, but I consider ok including the README which is only 35k

pitbulk:
Solved at https://github.com/onelogin/ruby-saml/commit/2e2e9b2a8e2074576dfc6e704b55ac5dc8171c15

PR #539, opened 2020-05-23
Expose which service binding was chosen for single sign on service

lyang:
# What
Expose which service binding was chosen for `idp_sso_target_url`

# Why
Knowing the chosen binding helps application code to choose:
1) `redirect_to` when the chosen binding is `HTTP-Redirect`
2) `render :post_with_js` when the chosen binding is `HTTP-POST`

### DIFF
--- a/lib/onelogin/ruby-saml/idp_metadata_parser.rb
+++ b/lib/onelogin/ruby-saml/idp_metadata_parser.rb
@@ -210,6 +210,7 @@ module OneLogin
             :idp_entity_id => @entity_id,
             :name_identifier_format => idp_name_id_format,
             :idp_sso_target_url => single_signon_service_url(options),
+            :single_signon_service_binding => single_signon_service_binding(options[:sso_binding]),
             :idp_slo_target_url => single_logout_service_url(options),
             :idp_attribute_names => attribute_names,
             :idp_cert => nil,
--- a/lib/onelogin/ruby-saml/settings.rb
+++ b/lib/onelogin/ruby-saml/settings.rb
@@ -32,6 +32,7 @@ module OneLogin
       # IdP Data
       attr_accessor :idp_entity_id
       attr_accessor :idp_sso_target_url
+      attr_accessor :single_signon_service_binding
       attr_accessor :idp_slo_target_url
       attr_accessor :idp_cert
       attr_accessor :idp_cert_fingerprint
--- a/test/idp_metadata_parser_test.rb
+++ b/test/idp_metadata_parser_test.rb
@@ -25,6 +25,7 @@ class IdpMetadataParserTest < Minitest::Test
 
       assert_equal "https://hello.example.com/access/saml/idp.xml", settings.idp_entity_id
       assert_equal "https://hello.example.com/access/saml/login", settings.idp_sso_target_url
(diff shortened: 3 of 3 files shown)


pitbulk:
You can save a state with the RequestId as well as the binding used and send it in the RelayState, later recover it on the ruby app.

lyang:
Hi @pitbulk , thank you for taking a look.

My primary use case is for SP initiated flow.

When parsing IdP metadata, I could pass in both `HTTP-Redirect` and `HTTP-POST` as supported `sso_binding`, like this:
```ruby
settings = OneLogin::RubySaml::IdpMetadataParser.new.parse(metadata, sso_binding: [HTTP_REDIRECT, HTTP_POST])
```
Currently, the parser will pick the `idp_sso_target_url` from the first match, from the metadata xml.

But from the resulting `settings` object, it only has the `idp_sso_target_url` value.

After I create the request object, I want to know which binding was picked, so that I can proceed correctly, something like:
```ruby
@auth_request = OneLogin::RubySaml::Authrequest.new.create(settings)

case settings.single_signon_service_binding
when HTTP_REDIRECT
  redirect_to @auth_request
when HTTP_POST
  render :post_with_js
else
  # ...
end
```

Thanks!

pitbulk:
but are you aware that the [AuthNRequest](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/authrequest.rb) class right now does not consider that binding that are you storing and always prepare the AuthNRequest for HTTP-Redirect binding?

lyang:
Hi @pitbulk , would you mind clarifying:
> always prepare the AuthNRequest for HTTP-Redirect binding?

As far as I can tell, `OneLogin::RubySaml::Authrequest.new.create(settings)` returns the login URL, as a `String`.
Given the `settings.idp_sso_target_url` is `sso_binding` aware, the resulting login url should be `sso_binding` aware as well.
All four query params `SAMLRequest`, `RelayState`, `SigAlg` and `Signature` are sso_binding agnostic.

So I think one can simply use it in a form to auto post to the idp, for e.g.
```erb
<body onload="document.forms['login'].submit()">
  <form name="login" method="POST" action="<%= @login_url %>">
  <input type="submit">Login</input>
  </form>
</body>
```

Thanks!

pitbulk:
Yes, but if the IdP requires signed AuthNRequest, the Signature for POST binding needs to be embedded instead of a GET parameter Signature.

lyang:
Ah, that makes sense. Thanks @pitbulk!
Is there a plan to implement `HTTP-POST` binding anytime soon?

pitbulk:
Not currently in the Roadmap.

HTTP-Redirect is a binding that IdPs should implement. Are you connecting with an IdP not supporting it?

lyang:
I was testing against metadata from https://samltest.id/saml/idp.
It has 3 bindings for SSO login and the first one wasn't `HTTP-Redirect` (or `HTTP-POST`).
So I had to pass `sso_binding => [HTTP_REDIRECT]` to make it work.

Then I was just trying to get `HTTP-POST` to work as well, because it looked simple enough.
The project is at https://github.com/lyang/saml-proxy, if you are interested.

Thanks!

lyang:
Closing this now

PR #545, opened 2020-09-19
More specific error messages for request certificate validation

FabioCareMonkey:
When validations find that the SAML request certificate does not match idp certificate they report the generic error message "Response signature validation failed";
This PR adds a more specific message for this scenario.

additionally the 'soff' variable which is required when calling the "append_error" method was not passed down to the 'validate_document_with_cert' method. This is also fixed

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -863,7 +863,7 @@ module OneLogin
           valid = false
           expired = false
           idp_certs[:signing].each do |idp_cert|
-            valid = doc.validate_document_with_cert(idp_cert)
+            valid = doc.validate_document_with_cert(idp_cert, @soft)
             if valid
               if settings.security[:check_idp_cert_expiration]
                 if OneLogin::RubySaml::Utils.is_cert_expired(idp_cert)
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -863,8 +863,12 @@ module OneLogin
           valid = false
           expired = false
           idp_certs[:signing].each do |idp_cert|
-            valid = doc.validate_document_with_cert(idp_cert, @soft)
+            valid = doc.validate_document_with_cert(idp_cert, true)
             if valid
+              # required to reset errors as there could be issues with previous certificates
+              doc.reset_errors!
+              reset_errors!
+
               if settings.security[:check_idp_cert_expiration]
                 if OneLogin::RubySaml::Utils.is_cert_expired(idp_cert)
                   expired = true
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -862,27 +862,37 @@ module OneLogin
         else
           valid = false
           expired = false
+          accumulated_doc_errors = []
           idp_certs[:signing].each do |idp_cert|
             valid = doc.validate_document_with_cert(idp_cert, true)
             if valid
-              # required to reset errors as there could be issues with previous certificates
-              doc.reset_errors!
-              reset_errors!
-
               if settings.security[:check_idp_cert_expiration]
                 if OneLogin::RubySaml::Utils.is_cert_expired(idp_cert)
                   expired = true
                 end
               end
+
+              #at least one certificate is valid, clear accumulated errors
+              accumulated_doc_errors = []
               break
             end
+
+            accumulated_doc_errors += doc.errors
+
+            # required to reset errors so we can collect other certificate errors
+            doc.reset_errors!
+            reset_errors!
           end
+
           if expired
             error_msg = "IdP x509 certificate expired"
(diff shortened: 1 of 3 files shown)


lib/xml_security.rb
```diff
@@ -260,7 +260,7 @@ def validate_document_with_cert(idp_cert)
 
         # check saml response cert matches provided idp cert
         if idp_cert.to_pem != cert.to_pem
-          return false
+          return append_error("SAML response certificate does not match idp certificate", soft)
```
FabioCareMonkey, line 263: the `soft`variable (from line 258) was missing on the scope of this method

lib/xml_security.rb
```diff
@@ -241,7 +241,7 @@ def validate_document(idp_cert_fingerprint, soft = true, options = {})
       validate_signature(base64_cert, soft)
     end
 
-    def validate_document_with_cert(idp_cert)
+    def validate_document_with_cert(idp_cert, soft = true)
```
FabioCareMonkey, line 244: the `soft`variable (from line 258) was missing on the scope of this method

test/xml_security_test.rb
```diff
@@ -395,25 +395,51 @@ class XmlSecurityTest < Minitest::Test
     end
 
     describe '#validate_document_with_cert' do
+      let(:document_data) { read_response('response_with_signed_message_and_assertion.xml') }
+      let(:document) { OneLogin::RubySaml::Response.new(document_data).document }
+      let(:idp_cert) { OpenSSL::X509::Certificate.new(ruby_saml_cert_text) }
+      let(:fingerprint) { '4b68c453c7d994aad9025c99d5efcf566287fe8d' }
```
FabioCareMonkey, line 401: I moved these up a couple of levels to reuse them in the other tests in at this level

test/xml_security_test.rb
```diff
@@ -395,25 +395,51 @@ class XmlSecurityTest < Minitest::Test
...
-          it 'is valid' do
-            assert document.validate_document_with_cert(idp_cert), 'Document should be valid'
+      describe 'when response cert is invalid' do
+        let(:document_data) do
+          contents = read_response('response_with_signed_message_and_assertion.xml')
+          contents.sub(/<ds:X509Certificate>.*<\/ds:X509Certificate>/,
+                       "<ds:X509Certificate>an-invalid-certificate</ds:X509Certificate>")
+        end
+
+        it 'is not valid' do
```
FabioCareMonkey, line 424: this would fail on the current version of the gem (and it should not)

lib/onelogin/ruby-saml/response.rb
```diff
@@ -863,8 +863,12 @@ def validate_signature
           valid = false
           expired = false
           idp_certs[:signing].each do |idp_cert|
-            valid = doc.validate_document_with_cert(idp_cert)
+            valid = doc.validate_document_with_cert(idp_cert, true)
```
pitbulk, line 866: It should be used
```
valid = doc.validate_document_with_cert(idp_cert, @soft)
```
instead
```
 valid = doc.validate_document_with_cert(idp_cert, true)
```

lib/xml_security.rb
```diff
@@ -255,12 +255,12 @@ def validate_document_with_cert(idp_cert)
         begin
           cert = OpenSSL::X509::Certificate.new(cert_text)
         rescue OpenSSL::X509::CertificateError => _e
-          return append_error("Certificate Error", soft)
+          return append_error("Document certificate error", soft)
```
pitbulk, line 258: Please keep the "Certificate Error" message, maybe there are integrations already listening for such string.

lib/xml_security.rb
```diff
@@ -255,12 +255,12 @@ def validate_document_with_cert(idp_cert)
...
         end
 
         # check saml response cert matches provided idp cert
         if idp_cert.to_pem != cert.to_pem
-          return false
+          return append_error("Document certificate does not match idp certificate", soft)
```
pitbulk, line 263: In an environment where multiple idps cert are registered , that error gonna be registered for each validation executed.
In my opinion, this error should not be registered here because can't confuse the admin reviewing the logs.

I'm ok raising an error that none of the registered certs matched the certificate in the document which will require some refactor in the code,

lib/onelogin/ruby-saml/response.rb
FabioCareMonkey: we cannot use the `@soft` variable here, as it would raise if any of the certificates is invalid. this validation is required to be soft.

I will adjust the code as per your second comment. and using the soft variable to determine if it should raise after validating all certificates

lib/xml_security.rb
FabioCareMonkey: Please let me know what you think about the new chagnes

pitbulk:
Followed a similar approach at: https://github.com/onelogin/ruby-saml/commit/59b9ade14709ec37bedc489714c90b0f527ec6da

Issue #541, opened 2020-05-30
Support for AES-128-GCM decryption


flyingolivier:
We are trying to use `ruby-saml` to implement a Service Provider (SP) that talks to a Shibboleth Identity Provider (IDP). The IDP returns an encrypted assertion using the `AES-128-GCM` algorithm. `ruby-saml` is failing to decrypt the assertion and returns the following error message:

> NoMethodError (undefined method `[]' for nil:NilClass)

Reading [the code](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/utils.rb#L251-L258 ) it appears that CBC mode algoritms such as `DES-EDE3-CBC`, `AES-128-CBC`, `AES-192-CBC` and `AES-256-CBC` are suported by `ruby-saml` but the `AES-GCM` modes are not.

Shibboleth recently [changed its encryption defaults to favor AES-GCM for XML encryption](https://wiki.shibboleth.net/confluence/display/IDP4/GCMEncryption). Considering that Shibboleth is widely used, It is likely that more and more Service Providers will be facing this issue.

Further investigations show that the OpenSSL library used by this gem already supports CGM modes.

```ruby
 irb(main):065:0> OpenSSL::Cipher.ciphers.include?("aes-128-gcm")
=> true
```

My knowledge in encryption/decryption algorithms is limited but I will gladly provide more information to see this support added.

ahukkanen:
Here is the method in `ruby-saml` that does the decryption of the encrypted SAML message:
https://github.com/onelogin/ruby-saml/blob/7423f03618c0f78f072d31132375cb4e2d89b282/lib/onelogin/ruby-saml/utils.rb#L250-L273

Here is how you can modify that to support AES-128-GCM, AES-192-GCM and AES-256-GCM:

```ruby
OneLogin::RubySaml::Utils.class_eval do
  def self.retrieve_plaintext(cipher_text, symmetric_key, algorithm)
    case algorithm
      when 'http://www.w3.org/2001/04/xmlenc#tripledes-cbc' then cipher = OpenSSL::Cipher.new('DES-EDE3-CBC').decrypt
      when 'http://www.w3.org/2001/04/xmlenc#aes128-cbc' then cipher = OpenSSL::Cipher.new('AES-128-CBC').decrypt
      when 'http://www.w3.org/2001/04/xmlenc#aes192-cbc' then cipher = OpenSSL::Cipher.new('AES-192-CBC').decrypt
[...]

PR #557, opened 2020-11-30
Add support for AES-GCM decryption

silppuri:
This PR adds support for AES-128-GCM, AES-192-GCM, and AES-256-GCM encryption methods.

## Implementation
  The implementation ended up being exactly the one that @ahukkanen described in the issue comment: https://github.com/onelogin/ruby-saml/issues/541#issuecomment-714646399  

The AES-GCM encryption is formed as described in https://www.w3.org/TR/xmlenc-core1/#sec-AES-GCM
  The ciphertext in AES GCM consists of 96-bit Initialization Vector (IV) + encrypted octets + and 128-bit Authentication Tag (T). No padding is used.   So the ciphertext message is as following
```
IV (96bit) + data + T (128bit)
```

## Tests
The tests fixtures were formed like this:
1. Copy the encrypted assertion and key from an existing AES-CBC test fixture to `assertion.xml` and `key.txt`
2. Form the encrypted message:
```rb
plaintext_response = File.read("assertion.xml")
cipher = OpenSSL::Cipher.new("AES-128-GCM").encrypt
iv = Random.bytes(12)
key = File.read("key.txt")
key.force_encoding("ASCII-8BIT")
cipher.iv = iv
cipher.key = key
cipher_text = cipher.update(plaintext_response) + cipher.final
auth_tag = cipher.auth_tag

# Form the message as described in https://www.w3.org/TR/xmlenc-core1/#sec-AES-GCM
base_64_string = Base64.strict_encode64(iv + cipher_text + auth_tag)
File.write("encrypted-assertion.base64", base_64_string)
```
3. Copy the `encrypted-assertion.base64` to response fixture and base64 encode it.

A plaintext example of the fixture can be found here https://gist.github.com/silppuri/121f2099ac06dfca35e96c674e6135ea

**TODO**
- [x] Test this in a test environment with an IdP that uses AES-GCM encryption


silppuri:
I tested this in a testing environment with and IdP using the AES-128-GCM algorithm successfully.

silppuri:
@pitbulk any way I can help to get this forward?

PR #559, opened 2020-12-08
Parse & use SLO Response Location, fixes #556

### LINKED ISSUE #556: IdpMetadataParser doesn't read SLO ResponseLocation
Hi! Ran into a similar issue as this one: https://github.com/onelogin/python3-saml/issues/13. Basically when parsing the following XML config `settings.idp_slo_target_url` will be set to `foo` rather than `bar`.

```xml
<ns0:SingleLogoutService
  Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"
  Location="foo"
  ResponseLocation="bar"
/>
```

 Would you accept a PR that sets settings.idp_slo_target_url to the value on ResponseLocation if it's available? If so, are there cases where both the Location and ResponseLocation would need to be used?

Note: I'm not sure if me/my employer will be able to prioritise this, but wanted to check in first if a potential PR would be accepted.

Relevant code: https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/idp_metadata_parser.rb#L263-L273

JCB-K:
Hi,

Here's the PR that fixes #556, i.e. parsing the SLO response location set on the IDP's config for use in the LogoutResponse class. Test cases have been added for it. When the `ResponseLocation` attribute is set, the request will look like this:

`/nidp/saml2/slo_return?SAMLResponse=<samlp:LogoutResponse Destination='/nidp/saml2/slo_return'...`


JCB-K:
@pitbulk Any update on this? Happy to resolve any potential issues.

PR #561, opened 2020-12-29
add runtime dependency rexml

tknzk:
rexml are rexml are default gems to bundled gem for ruby 2.8 (3.0)
https://bugs.ruby-lang.org/issues/16485


fwininger:
thanks ! :+1:

pitbulk:
Covered with https://github.com/onelogin/ruby-saml/pull/562/files

Issue #409, opened 2017-07-15
idp_cert_multi issue w/ the hash storing the keys


antislice:
I ran into an issue getting with getting "Invalid Signature on SAML Response" on a response with with `idp_cert_multi` set. The IdP is ADFS if anyone's searching. I tracked it down to a hash access keys.

`IdpMetadataParser#certificates` uses string keys to save the keys, `'signing'` and `'encryption'`. This is a problem in `Settings#get_idp_cert_multi` when it tries to access the cert with the keys `:signing` and `:encryption`. In plain Ruby without ActiveSupport's HashWithIndifferentAccess the two forms are not interchangeable.

I was able to work around it with the two lines after `parse_remote` when I initialize the settings:
```
  def self.get_saml_settings(url_base)
    # retrieve settings from IdP 
    unless @settings
      idp_metadata_parser = OneLogin::RubySaml::IdpMetadataParser.new
      @settings = idp_metadata_parser.parse_remote(ENV['SSO_IDP_METADATA_URL'])
      # save the parts of the cert with symbol keys
      @settings.idp_cert_multi[:signing] = @settings.idp_cert_multi['signing']
      @settings.idp_cert_multi[:encryption] = @settings.idp_cert_multi['encryption']
  end
  ...
  # the rest of the settings
end
```

pitbulk:
We may change the way we calculate the [certificates](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/idp_metadata_parser.rb#L250) and use the symbol keys.
And update also the [merge_certificates_into](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/idp_metadata_parser.rb#L318) method.

Do you want to contribute with a PR?

antislice:
Yeah, I'm planning on it when I get the chance. This came up in the middle of a lot of deadlines so I wasn't able to submit a PR with the issue. It stumped me long enough I wanted to get the workaround out though.

PR #576, opened 2021-03-10
Support idp cert multi with string keys

alxckn:
I ran into issues while attempting to integrate with an ADFS IdP.
The issue was that when settings are initialized with a hash where keys are not symbols (ex. when loaded from a json file), `Settings#get_idp_cert_multi` is not able to retrieve the certificates. 

It had been reported in this issue before: https://github.com/onelogin/ruby-saml/issues/409.

This PR proposes to fallback to string keys if symbol keys are not defined.

### DIFF
--- a/lib/onelogin/ruby-saml/settings.rb
+++ b/lib/onelogin/ruby-saml/settings.rb
@@ -195,17 +195,13 @@ module OneLogin
 
         certs = {:signing => [], :encryption => [] }
 
-        if idp_cert_multi.key?(:signing) and not idp_cert_multi[:signing].empty?
-          idp_cert_multi[:signing].each do |idp_cert|
-            formatted_cert = OneLogin::RubySaml::Utils.format_cert(idp_cert)
-            certs[:signing].push(OpenSSL::X509::Certificate.new(formatted_cert))
-          end
-        end
+        [:signing, :encryption].each do |type|
+          certs_for_type = idp_cert_multi[type] || idp_cert_multi[type.to_s]
+          next if !certs_for_type || certs_for_type.empty?
 
-        if idp_cert_multi.key?(:encryption) and not idp_cert_multi[:encryption].empty?
-          idp_cert_multi[:encryption].each do |idp_cert|
+          certs_for_type.each do |idp_cert|
             formatted_cert = OneLogin::RubySaml::Utils.format_cert(idp_cert)
-            certs[:encryption].push(OpenSSL::X509::Certificate.new(formatted_cert))
+            certs[type].push(OpenSSL::X509::Certificate.new(formatted_cert))
           end
         end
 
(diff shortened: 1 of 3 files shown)


johnnyshields:
👍 This looks good to merge

lib/onelogin/ruby-saml/settings.rb
```diff
@@ -198,17 +198,13 @@ def get_idp_cert_multi
...
+        [:signing, :encryption].each do |type|
+          certs_for_type = idp_cert_multi[type] || idp_cert_multi[type.to_s]
+          next if !certs_for_type || certs_for_type.empty?
 
-        if idp_cert_multi.key?(:encryption) and not idp_cert_multi[:encryption].empty?
-          idp_cert_multi[:encryption].each do |idp_cert|
+          certs_for_type.each do |idp_cert|
             formatted_cert = OneLogin::RubySaml::Utils.format_cert(idp_cert)
-            certs[:encryption].push(OpenSSL::X509::Certificate.new(formatted_cert))
+            certs[type].push(OpenSSL::X509::Certificate.new(formatted_cert))
```
pitbulk, line 207: It should be:
```
certs[type.to_sym].push(OpenSSL::X509::Certificate.new(formatted_cert))
```

So we always generate certs with the symbol index, as it used to be.

test/response_test.rb
```diff
@@ -972,6 +972,16 @@ def generate_audience_error(expected, actual)
         assert_empty response_valid_signed.errors
       end
 
+      it "return true when at least a cert on idp_cert_multi is valid and keys are strings" do
+        settings.idp_cert_multi = {
+          "signing" => [ruby_saml_cert_text2, ruby_saml_cert_text],
+          "encryption" => []
+        }
+        response_valid_signed.settings = settings
+        res = response_valid_signed.send(:validate_signature)
```
pitbulk, line 981: In addition to this unittest,  I expect one using a call to get_idp_cert_multi as well that verifies the certs were properly loaded on the settings object at settings_test.rb

alxckn: Added tests verifying that calls to get_idp_cert_multi are in fact working: 01c95db

lib/onelogin/ruby-saml/settings.rb
alxckn: `type` is a symbol, it can only be one of `[:signing, :encryption]`

alxckn:
Thanks for the review @johnnyshields, @pitbulk 
Sorry for the late changes

lib/onelogin/ruby-saml/settings.rb
johnnyshields: @alxckn 's code is correct here.

johnnyshields:
@alxckn this looks ok, however, I wonder if we should coerce the settings keys to be symbols when reading in the data.

@pitbulk would suggest to merge this as-is.

pitbulk (approved):

pitbulk (approved):

PR #583, opened 2021-04-30
000/downcase algorithm

zeos90:


pitbulk:
The algorithm issue was fixed in the latest release:
https://github.com/onelogin/ruby-saml/commit/fba4ed018e1010f7570d39d3f927adaf24a6a689

but is missed on master

PR #586, opened 2021-05-10
Support milliseconds in cacheDuration parsing

urbanautomaton:
Since the introduction of `cacheDuration` parsing in 25cbddd we've been seeing metadata parsing failures for one of our IdPs, whose `cacheDuration` value is set to `"PT6H0M0.000S"`

This seems like a valid ISO8601 duration, however the regexp being used for parsing doesn't allow for the possibility of milliseconds.

This PR adds an optional non-captured group to the regexp to permit them. Since the captured string value is already being converted using `.to_f`, this should work without further changes.

To achieve minimal test coverage I've added a seconds string to one of the existing metadata examples with a zero value; this reproduces the failure and confirms the fix without needing to add a specific test case.

### Feedback

I'm new to this project so absolutely any feedback would be welcome. In particular, is the approach taken to test coverage sufficient, or would you prefer to add an explicit test case? This seemed simpler since the duration parsing isn't directly tested, but I'm happy to introduce unit tests there if that's preferred.

### DIFF
--- a/lib/onelogin/ruby-saml/utils.rb
+++ b/lib/onelogin/ruby-saml/utils.rb
@@ -16,18 +16,18 @@ module OneLogin
       DSIG      = "http://www.w3.org/2000/09/xmldsig#"
       XENC      = "http://www.w3.org/2001/04/xmlenc#"
       DURATION_FORMAT = %r(^
-        (-?)P                         # 1: Duration sign
+        (-?)P                       # 1: Duration sign
         (?:
-          (?:(\d+)Y)?                 # 2: Years
-          (?:(\d+)M)?                 # 3: Months
-          (?:(\d+)D)?                 # 4: Days
+          (?:(\d+)Y)?               # 2: Years
+          (?:(\d+)M)?               # 3: Months
+          (?:(\d+)D)?               # 4: Days
           (?:T
-            (?:(\d+)H)?               # 5: Hours
-            (?:(\d+)M)?               # 6: Minutes
-            (?:(\d+)S)?   # 7: Seconds and optional milliseconds
+            (?:(\d+)H)?             # 5: Hours
+            (?:(\d+)M)?             # 6: Minutes
+            (?:(\d+(?:[.,]\d+)?)S)? # 7: Seconds
           )?
           |
-          (\d+)W                      # 8: Weeks
+          (\d+)W                    # 8: Weeks
         )
       $)x
 
@@ -59,13 +59,8 @@ module OneLogin
           raise Exception.new("Invalid ISO 8601 duration")
         end
 
-        durYears = matches[2].to_i
-        durMonths = matches[3].to_i
-        durDays = matches[4].to_i
-        durHours = matches[5].to_i
-        durMinutes = matches[6].to_i
-        durSeconds = matches[7].to_f
-        durWeeks = matches[8].to_i
+        durYears, durMonths, durDays, durHours, durMinutes, durSeconds, durWeeks =
+          matches[2..8].map { |match| match ? match.tr(',', '.').to_f : 0.0 }
 
         if matches[1] == "-"
           durYears = -durYears
--- a/lib/onelogin/ruby-saml/utils.rb
+++ b/lib/onelogin/ruby-saml/utils.rb
@@ -15,7 +15,21 @@ module OneLogin
 
       DSIG      = "http://www.w3.org/2000/09/xmldsig#"
       XENC      = "http://www.w3.org/2001/04/xmlenc#"
-      DURATION_FORMAT = %r(^(-?)P(?:(?:(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?)|(?:(\d+)W))$)
+      DURATION_FORMAT = %r(^
+        (-?)P                         # 1: Duration sign
+        (?:
+          (?:(\d+)Y)?                 # 2: Years
+          (?:(\d+)M)?                 # 3: Months
+          (?:(\d+)D)?                 # 4: Days
+          (?:T
+            (?:(\d+)H)?               # 5: Hours
+            (?:(\d+)M)?               # 6: Minutes
+            (?:(\d+)S)?   # 7: Seconds and optional milliseconds
+          )?
+          |
+          (\d+)W                      # 8: Weeks
+        )
+      $)x
 
       # Checks if the x509 cert provided is expired
       #
@@ -37,10 +51,10 @@ module OneLogin
       #                            current time.
       #
       # @return [Integer] The new timestamp, after the duration is applied.
-      # 
+      #
       def self.parse_duration(duration, timestamp=Time.now.utc)
         matches = duration.match(DURATION_FORMAT)
-      
+
         if matches.nil?
           raise Exception.new("Invalid ISO 8601 duration")
         end
--- a/lib/onelogin/ruby-saml/utils.rb
+++ b/lib/onelogin/ruby-saml/utils.rb
@@ -59,18 +59,10 @@ module OneLogin
           raise Exception.new("Invalid ISO 8601 duration")
         end
 
-        durYears, durMonths, durDays, durHours, durMinutes, durSeconds, durWeeks =
-          matches[2..8].map { |match| match ? match.tr(',', '.').to_f : 0.0 }
+        sign = matches[1] == '-' ? -1 : 1
 
-        if matches[1] == "-"
-          durYears = -durYears
-          durMonths = -durMonths
-          durDays = -durDays
-          durHours = -durHours
-          durMinutes = -durMinutes
-          durSeconds = -durSeconds
-          durWeeks = -durWeeks
-        end
+        durYears, durMonths, durDays, durHours, durMinutes, durSeconds, durWeeks =
+          matches[2..8].map { |match| match ? sign * match.tr(',', '.').to_f : 0.0 }
 
         initial_datetime = Time.at(timestamp).utc.to_datetime
         final_datetime = initial_datetime.next_year(durYears)
--- a/test/utils_test.rb
+++ b/test/utils_test.rb
@@ -1,6 +1,41 @@
 require File.expand_path(File.join(File.dirname(__FILE__), "test_helper"))
 
 class UtilsTest < Minitest::Test
+  describe ".parse_duration" do
+    DURATIONS_FROM_EPOCH = {
+      # Basic formats
+      "P1Y1M1D"        => "1971-02-02T00:00:00.000Z",
+      "PT1H1M1S"       => "1970-01-01T01:01:01.000Z",
+      "P1W"            => "1970-01-08T00:00:00.000Z",
+      "P1Y1M1DT1H1M1S" => "1971-02-02T01:01:01.000Z",
+
+      # Negative duration
+      "-P1Y1M1DT1H1M1S" => "1968-11-29T22:58:59.000Z",
+
+      # Nominal wraparounds
+      "P13M" => "1971-02-01T00:00:00.000Z",
+      "P31D" => "1970-02-01T00:00:00.000Z",
+    }
+
+    def result(duration, reference = 0)
+      Time.at(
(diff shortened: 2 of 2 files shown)


urbanautomaton:
Somewhat cheekily I've pushed another commit that reformats the regexp for readability. I was restricted in what I could do since this gem supports ruby 1.8.7, which doesn't have named capture groups. I think the split lines and comments make it a lot easier to understand and work with, but I'm more than happy to get rid of that commit if it's not preferred.

I've ensured the build passes with ruby 2.6.6, but I can't install ruby 1.8.7 on my mac any more. Is there any CI in place for this gem that tests a range of versions?

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -15,7 +15,21 @@ class Utils
...
+          (?:(\d+)D)?                 # 4: Days
+          (?:T
+            (?:(\d+)H)?               # 5: Hours
+            (?:(\d+)M)?               # 6: Minutes
+            (?:(\d+(?:\.\d{3})?)S)?   # 7: Seconds and optional milliseconds
+          )?
+          |
+          (\d+)W                      # 8: Weeks
+        )
+      $)x
```
pitbulk, line 32: what is this x?

pitbulk:
We use travis. but it seems the travis.org was frozen, so we need to move to travis.com
https://travis-ci.org/github/onelogin/ruby-saml/builds

lib/onelogin/ruby-saml/utils.rb
urbanautomaton: That's the "ignore whitespace" regexp flag. It's what allows me to split the regexp over multiple lines, and also insert comments. You can see it here documented for 2.6, but it's supported back to 1.8.x:

https://ruby-doc.org/core-2.6.4/Regexp.html#class-Regexp-label-Options

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -15,7 +15,21 @@ class Utils
...
+      DURATION_FORMAT = %r(^
+        (-?)P                         # 1: Duration sign
+        (?:
+          (?:(\d+)Y)?                 # 2: Years
+          (?:(\d+)M)?                 # 3: Months
+          (?:(\d+)D)?                 # 4: Days
+          (?:T
+            (?:(\d+)H)?               # 5: Hours
+            (?:(\d+)M)?               # 6: Minutes
+            (?:(\d+(?:\.\d{3})?)S)?   # 7: Seconds and optional milliseconds
```
johnnyshields, line 27: Should we support `{1,6}` decimal places?

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -15,7 +15,21 @@ class Utils
 
       DSIG      = "http://www.w3.org/2000/09/xmldsig#"
       XENC      = "http://www.w3.org/2001/04/xmlenc#"
-      DURATION_FORMAT = %r(^(-?)P(?:(?:(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?)|(?:(\d+)W))$)
+      DURATION_FORMAT = %r(^
+        (-?)P                         # 1: Duration sign
+        (?:
+          (?:(\d+)Y)?                 # 2: Years
```
johnnyshields, line 21: According to [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601) all values (years, months, etc.) may have fractions.

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -15,7 +15,21 @@ class Utils
...
+          (?:(\d+)M)?                 # 3: Months
+          (?:(\d+)D)?                 # 4: Days
```
johnnyshields, line 23: Shouldn't weeks (W) go before days (D)?

urbanautomaton:
Thanks for the notes, @johnnyshields! All of your points are good ones, but I'm a little wary of expanding the scope of this PR to attempt full coverage of the ISO8601 duration spec, given the indirect test coverage we've got to work with. Since two independent PRs have been raised for troubles with milliseconds specifically, I'd probably favour merging this to begin with and looking at full coverage as a follow-up.

I did have a quick look for libraries that could do this whole task for us, but unfortunately the most commonly-used one (outside of ActiveSupport), [ruby-duration](https://github.com/josedonizetti/ruby-duration), requires ruby 1.9, while ruby-saml still supports 1.8.7. They _do_ point to the [last version that supported earlier rubies](https://github.com/josedonizetti/ruby-duration/tree/v2.1.4), though, so we could potentially do another conditional dependency in the gemspec.

@pitbulk what do you think? Are you open to pulling in an extra dependency for this task, or would you prefer to continue updating this gem's implementation?

lib/onelogin/ruby-saml/utils.rb
urbanautomaton: From [the spec](https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf), 4.4.3.2:

> In	both	basic and extended format the complete representation of the expression for duration shall be [PnnW] or [PnnYnnMnnDTnnHnnMnnS].

So I think the current week-handling is correct.

lib/onelogin/ruby-saml/utils.rb
urbanautomaton: In [S4.4.3.2](https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf):

> If necessary for a particular application, the lowest order components may have a decimal fraction.

Which is vague, but I think means that only the H/M/S components may be fractional?

urbanautomaton:
In fact, after a bit of spec poking I think we actually weren't so far from having full spec coverage, so I've added a couple of commits to do that; we shouldn't need another dependency.

I've also rebased to pick up the new github actions build.

lib/onelogin/ruby-saml/utils.rb
johnnyshields: Hmmm... you are correct regarding the spec, however following the [https://en.wikipedia.org/wiki/Robustness_principle], since this a reading method we should still add it inline, as there doesn't seem to be any harm in doing so (weeks and years/months/days/etc. can coexist without conflict in the same string--just add them together).

lib/onelogin/ruby-saml/utils.rb
johnnyshields: "Lowest order" means the smallest value in the string. So `P0.5Y` or `P1Y0.5M` are both valid.

However, following the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle), and for code cleanliness reasons, let's just support decimal on all values, since there is no harm/ambiguity in doing so.

urbanautomaton: I wondered about that interpretation, but the plural in "lowest order components" makes me doubt it. I also noted in the commit message that the `DateTime#next_year` / `#next_month` and `#next_day` methods that we're using to compute the duration for these components ignore any fractional parts, so even if we parse these components as decimals, we can't compute their value without a lot more changes.

I'm also not certain that fractional increments of a nominal duration (Y/M/W/D) are easily interpretable. For example, `DateTime#next_month` computes the same day of the month, N calendar months in the future. This has a clearly defined meaning, and there are still complications due to different-length months. E.g. Jan 31 + 1 month = Feb 28 (or Feb 29 on a leap year). What should the result be if we added half a month instead? Damned if I know. 😄

So unless we genuinely support fractional years/months/days, we're not following the robustness principle, and there is a harm in parsing them: clients sending e.g. `P0.5Y` will get a duration of zero instead of an informative error.

lib/onelogin/ruby-saml/utils.rb
urbanautomaton: Since this is one of the few areas where the spec is mercifully clear, and the code prior to this PR already reflected the spec, I'm not keen to change it. We can always open another PR if there's a genuine need to support invalid duration strings, but I'd like to keep this one focused.

johnnyshields: > ...programs that receive messages should accept non-conformant input as long as the meaning is clear.

The meaning is 100% clear here if weeks are given alongside years, months, and days.

urbanautomaton: To repeat: I'd like to focus this PR on ensuring that we accept as many ISO8601-valid durations as possible. If you'd like to pursue support for invalid durations, by all means open another PR.

lib/onelogin/ruby-saml/utils.rb
johnnyshields: Yeah seems reasonable to support fractional Hour, Minute, and Day as the others are non-deterministic. Even Day and Week can be different than 24h on Daylight Saving Time transitions, so there is potential ambiguity which is better to reject.

That being said, we should allow each one of H/M/S to have a fraction `PT0.5H0.5M0.1S`

lib/onelogin/ruby-saml/utils.rb
johnnyshields: Fair enough. Let's get this one merged then I'll open another one :)

lib/onelogin/ruby-saml/utils.rb
urbanautomaton: Thanks - yep, I think the code as written will pick up the fractions on any or all of those components.

pitbulk (approved):

pitbulk:
Can we add more unitest to cover the different scenarios? maybe you can carry them from the gem dependency you found

urbanautomaton:
@pitbulk Sure, yep, I've:

1) Added unit tests in an initial commit to cover the pre-existing behaviour
2) Rewritten the regexp for clarity second
3) Then added full H/M/S decimal support in a single commit

The tests pass at each stage.

I didn't borrow the examples from the other gem in the end, because our method applies the duration to a timestamp as well as computing it whereas theirs produces a duration object, but I think I've gotten reasonable coverage of the major cases we want to support. See what you think!

test/utils_test.rb
```diff
@@ -1,6 +1,49 @@
...
+  describe ".parse_duration" do
+    DURATIONS_FROM_EPOCH = {
+      # Basic formats
+      "P1Y1M1D"        => "1971-02-02T00:00:00.000Z",
+      "PT1H1M1S"       => "1970-01-01T01:01:01.000Z",
+      "P1W"            => "1970-01-08T00:00:00.000Z",
+      "P1Y1M1DT1H1M1S" => "1971-02-02T01:01:01.000Z",
+
+      # Negative duration
+      "-P1Y1M1DT1H1M1S" => "1968-11-29T22:58:59.000Z",
```
johnnyshields, line 13: No issue support negative duration, but it's not clear what the entire duration being negative would mean in the context of `cacheDuration`

Also, are negative values supported for each unit, e.g. `P1Y-1M` == 11 months?

urbanautomaton: I'm not certain that negative durations are part of the spec, as it mentions them only once, to say:

> **2.1.6**
> duration
> non-negative quantity attributed to a time interval

From conversations in the moment.js repo ([1](https://github.com/moment/moment/issues/2408), [2](https://github.com/moment/moment/issues/3960)) it appears that this is an area where implementations vary, some supporting prefix signs, some supporting infix.

For the purposes of this PR, I'm just trying to cover the existing code, which supports a prefix sign alone.

pitbulk: Some info about the xsd:duration type: http://www.datypic.com/sc/xsd/t-xsd_duration.html

It seems infix should not be valid

urbanautomaton: Ah, cool, thanks for the extra info!

As a result of the earlier conversations I've permitted decimal hours and minutes; would you like me to back that change out, since the xsd:duration spec is tighter than the ISO one?

pitbulk: Yes please

urbanautomaton: No worries, glad to have it resolved! I've updated the branch and referred to the xsd:duration docs in the commit message.

johnnyshields:
Let's merge?

PR #589, opened 2021-05-14
Added Regex support for fractional seconds duration

gpinkham:
## Status
**READY**

## Migrations
NO

## Description
Had a customer send in XML that had a valid duration with fractional seconds  (0.000S) but the Regex parser did not support that. .  Updated Regex expression to support fractional seconds.    Backwards compatible to support integer seconds (0S)

## Related PRs
None

## Todos
- [X ] Tests

## Deploy Notes
None

## Steps to Test or Reproduce
Duration that failed with current code: "P0Y0M30DT0H0M0.000S"

* Utils#parse_duration

### DIFF
--- a/lib/onelogin/ruby-saml/utils.rb
+++ b/lib/onelogin/ruby-saml/utils.rb
@@ -15,7 +15,7 @@ module OneLogin
 
       DSIG      = "http://www.w3.org/2000/09/xmldsig#"
       XENC      = "http://www.w3.org/2001/04/xmlenc#"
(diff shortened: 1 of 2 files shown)


gpinkham:
just realized someone else already has a PR for this fix..     only diff is the way we tested   I'll leave this here just in case it's needed but feel free to close it if you want..     thanks!

johnnyshields:
Think we should close this in favor of https://github.com/onelogin/ruby-saml/pull/586

PR #594, opened 2021-05-25
Add warning about the use of IdpMetadataParser class.

pitbulk:
If Metadata URLs are provided by 3rd parties, the URL inputs MUST be validated to avoid issues like SSRF


gwadley (commented):
Suggested README Update:

Usually the same administrator that handles the Service Provider also sets the URL to the IdP, which should be a trusted resource.

But there are other scenarios, like a SAAS app where the administrator of the app delegates this functionality to other users. In this case, extra precaution should be taken in order to validate such URL inputs and avoid attacks like SSRF.

luisvm (approved):
might need a sign off from @gwadley as well 👍

PR #597, opened 2021-06-01
If URI='' use document root

zeos90:
URI=""
  Identifies the node-set (minus any comment nodes) of the XML resource containing the signature

https://www.w3.org/TR/xmldsig-core1/


zeos90:
Hi,

  Anything I need to do to get this considered? I am open to make changes/improvements if required.

Thanks,

pitbulk:
On this thread Scott Cantor says that empty URIs shouldn't be allowed in SAML implementations
http://shibboleth.net/pipermail/dev/2016-April/008162.html

So not implementing this here.

Issue #598, opened 2021-06-22
Make max byte size configurable


dblessing:
https://github.com/onelogin/ruby-saml/commit/533c84ebfc40f8cbac645b6c76ce4949f95d27d6 mitigated a vulnerability by creating a static max byte size (related to https://github.com/onelogin/ruby-saml/pull/383). We've encountered some cases where the SAML Response exceeds this size due to custom claims, such as when including groups a user is a member of. 

For this reason, it would be great if the max byte size was configurable. 

@pitbulk Do you think this would be a good feature?

pitbulk:
What byte size has your SAMLResponse?

dblessing:
@pitbulk In one case it looks like it was just less than .6MB. We believe some are even larger. 

If it's configurable then each instance can increase it if necessary - otherwise it remains the low .25MB default.

PR #601, opened 2021-07-15
Replace MAX_BYTE_SIZE constant with setting

gitlab-djensen:
## Status
**READY**

## Migrations
NO

## Description
The MAX_BYTE_SIZE constant did not allow for customization, which
is necessary for cases where legitimate SAML responses are larger
than 250,000 bytes. This replaces the constant with a setting, which
has a default value of 250,000 bytes, but can be customized like
any other setting.

This was motivated by https://github.com/onelogin/ruby-saml/issues/598, which was motivated by https://gitlab.com/gitlab-org/gitlab/-/issues/329053.
This follows the [existing pattern used for `encode_raw_saml`](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/saml_message.rb#L111).

## Related PRs
List related PRs against other branches:

branch | PR
------ | ------
Introduce MAX_BYTE_SIZE | [link](https://github.com/onelogin/ruby-saml/pull/566/files#diff-6e890879e7944b6cac0e115f001b790f412dc5f6a9a4a11651f70f31c6a33124R25)
Introduce inflate option to disable SAML message inflation | [link](https://github.com/onelogin/ruby-saml/pull/383)

## Todos
- [x] Tests
- [x] Documentation

## Deploy Notes
Nothing noteworthy.

## Steps to Test or Reproduce
Outline the steps to test or reproduce the PR here.

```sh
git fetch
git checkout gitlab-djensen:gitlab-djensen-use-setting-for-max-byte-size
bundle install;
ruby -Itest test/settings_test.rb
```

* Settings
* SamlMessage

### DIFF
--- a/lib/onelogin/ruby-saml/saml_message.rb
+++ b/lib/onelogin/ruby-saml/saml_message.rb
@@ -22,8 +22,6 @@ module OneLogin
       BASE64_FORMAT = %r(\A([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?\Z)
       @@mutex = Mutex.new
 
-      MAX_BYTE_SIZE = 250000
-
       # @return [Nokogiri::XML::Schema] Gets the schema object of the SAML 2.0 Protocol schema
       #
       def self.schema
@@ -88,11 +86,12 @@ module OneLogin
       # @param saml [String] The deflated and encoded SAML Message
       # @return [String] The plain SAML Message
       #
-      def decode_raw_saml(saml)
+      def decode_raw_saml(saml, settings = nil)
         return saml unless base64_encoded?(saml)
 
-        if saml.bytesize > MAX_BYTE_SIZE
-          raise ValidationError.new("Encoded SAML Message exceeds " + MAX_BYTE_SIZE.to_s + " bytes, so was rejected")
+        settings = OneLogin::RubySaml::Settings.new if settings.nil?
+        if saml.bytesize > settings.message_max_bytesize
+          raise ValidationError.new("Encoded SAML Message exceeds " + settings.message_max_bytesize.to_s + " bytes, so was rejected")
         end
 
         decoded = decode(saml)
--- a/lib/onelogin/ruby-saml/settings.rb
+++ b/lib/onelogin/ruby-saml/settings.rb
@@ -54,6 +54,7 @@ module OneLogin
       attr_accessor :compress_request
       attr_accessor :compress_response
       attr_accessor :double_quote_xml_attribute_values
+      attr_accessor :message_max_bytesize
       attr_accessor :passive
       attr_reader   :protocol_binding
       attr_accessor :attributes_index
@@ -264,6 +265,7 @@ module OneLogin
         :idp_cert_fingerprint_algorithm            => XMLSecurity::Document::SHA1,
         :compress_request                          => true,
         :compress_response                         => true,
+        :message_max_bytesize                      => 250000,
         :soft                                      => true,
         :double_quote_xml_attribute_values         => false,
         :security                                  => {
--- a/README.md
+++ b/README.md
@@ -816,6 +816,27 @@ response = OneLogin::RubySaml::Response.new(params[:SAMLResponse], :allowed_cloc
 
 Make sure to keep the value as comfortably small as possible to keep security risks to a minimum.
 
+## Deflation Limit
+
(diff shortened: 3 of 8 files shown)


gitlab-djensen:
@dblessing would you mind doing the initial review on this whenever you have some time?

dblessing:
@gitlab-djensen Thanks for opening the PR. Looks good to me, with my limited knowledge of the `ruby-saml` code. Let's see what the maintainers think.

gitlab-djensen:
@pitbulk this PR is ready for your review, whenever you might have some time.

GitLab is hoping you will release a new version of this gem containing this fix, because `MAX_BYTE_SIZE` was a breaking change for implementations with large messages ([example](https://gitlab.com/gitlab-org/gitlab/-/issues/329053)).

johnnyshields:
👍 looks good to merge

gitlab-djensen:
@pitbulk friendly ping, hopefully this is an easy review 🙂

lib/onelogin/ruby-saml/saml_message.rb
```diff
@@ -88,11 +88,11 @@ def valid_saml?(document, soft = true)
       # @param saml [String] The deflated and encoded SAML Message
       # @return [String] The plain SAML Message
       #
-      def decode_raw_saml(saml)
+      def decode_raw_saml(saml, settings)
```
pitbulk, line 91: instead the whole setting object, can we just add message_max_bytesize param by default MAX_BYTE_SIZE?

gitlab-djensen: @pitbulk I'm pretty sure we could, but this follows the [existing pattern used for `encode_raw_saml`](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/saml_message.rb#L111) (the inverse method).  Do you want me to change this method even though it would then differ from `encode_raw_saml`?

pitbulk: ohh you are right, let's then keep it as it is

lib/onelogin/ruby-saml/settings.rb
```diff
@@ -258,6 +259,7 @@ def get_sp_key
         :idp_cert_fingerprint_algorithm            => XMLSecurity::Document::SHA1,
         :compress_request                          => true,
         :compress_response                         => true,
+        :message_max_bytesize                      => 250000,
```
pitbulk, line 262: Can this default value be read from SamlMessage.MAX_BYTE_SIZE?

gitlab-djensen: @pitbulk ha, I missed that! Just pushed a commit that removes it 👍

gitlab-djensen:
@pitbulk thanks for reviewing!  I just marked both of your comments as resolved.  Should we run the workflow next?  (It's telling me "First-time contributors need a maintainer to approve running workflows.")

gitlab-djensen:
Strangely, [the workflow ran anyway](https://github.com/onelogin/ruby-saml/actions/runs/1125004867).  There were a bunch of failures because `settings` was still optional in some places where it is now required.  I just pushed 2 commits that make `settings` a required argument in multiple places, **which makes this PR a breaking change**.   (We could avoid the breaking change by leaving `settings` as an optional argument and calling `OneLogin::RubySaml::Settings.new` when `settings` is not provided.  But that approach is not currently used anywhere in the gem, so it would be a new pattern.)  I'll keep working on this until the tests are green.

pitbulk:
We need to avoid breaking changes. If settings is not provided to the decode_raw_saml method, makes sense to keep using the default value, otherwise, the settings object should be there.

gitlab-djensen:
@pitbulk gotcha, refactored to be non-breaking (by calling `OneLogin::RubySaml::Settings.new` when `settings` is not passed).  Tests are green.  How's it looking now?

johnnyshields:
@pitbulk this looks good to me.

gitlab-djensen:
@pitbulk friendly ping, [prompted by a colleague](https://gitlab.com/gitlab-org/gitlab/-/issues/329053#note_667767935) with an affected client.  Does this look good for merge?

johnnyshields:
Also, after merge, let's get an gem release please :) Lots of great improvements recently.

pitbulk:
Sorry I will be back from PTO on 6 and take care of all the pending PRs

pitbulk (approved):

PR #602, opened 2021-08-07
Refactor the OneLogin::RubySaml::Metadata class

### LINKED ISSUE #433: enable additional SP metadata
My idp requires me to add additional fields (e.g. Organization, ContactPerson, ...) to my metadata before I can register my SP with them.

Is there a way to add these to the generated metadata?

johnnyshields:
Fixes #433

- Refactor the OneLogin::RubySaml::Metadata class so it is easier to extend by breaking it into a series of smaller methods.
- Also adds the #add_extras convenience method which is empty but can be extended.
- No change to behavior or to public method surface area.

Example of extending the class:

```ruby
  class MyMetadata < OneLogin::RubySaml::Metadata
    def add_extras(root, _settings)
      org = root.add_element("md:Organization")
      org.add_element("md:OrganizationName", 'xml:lang' => "en-US").text = 'ACME Inc.'
      org.add_element("md:OrganizationDisplayName", 'xml:lang' => "en-US").text = 'ACME'
      org.add_element("md:OrganizationURL", 'xml:lang' => "en-US").text = 'https://www.acme.com'

      cp = root.add_element("md:ContactPerson", 'contactType' => 'technical')
      cp.add_element("md:GivenName").text = 'ACME SAML Team'
      cp.add_element("md:EmailAddress").text = 'saml@acme.com'
    end
  end
```

### DIFF
--- a/lib/onelogin/ruby-saml/metadata.rb
+++ b/lib/onelogin/ruby-saml/metadata.rb
@@ -21,21 +21,50 @@ module OneLogin
       #
       def generate(settings, pretty_print=false, valid_until=nil, cache_duration=nil)
         meta_doc = XMLSecurity::Document.new
+        add_xml_declaration(meta_doc)
+        root = add_root_element(meta_doc, settings, valid_until, cache_duration)
+        sp_sso = add_sp_sso_element(root, settings)
+        add_sp_certificates(sp_sso, settings)
+        add_sp_service_elements(sp_sso, settings)
+        add_extras(root, settings)
+        embed_signature(meta_doc, settings)
+        output_xml(meta_doc, pretty_print)
+      end
+
+      protected
+
+      def add_xml_declaration(meta_doc)
+        meta_doc << REXML::XMLDecl.new('1.0', 'UTF-8')
+      end
+
+      def add_root_element(meta_doc, settings, valid_until, cache_duration)
         namespaces = {
             "xmlns:md" => "urn:oasis:names:tc:SAML:2.0:metadata"
         }
+
         if settings.attribute_consuming_service.configured?
           namespaces["xmlns:saml"] = "urn:oasis:names:tc:SAML:2.0:assertion"
         end
-        root = meta_doc.add_element "md:EntityDescriptor", namespaces
-        sp_sso = root.add_element "md:SPSSODescriptor", {
+
+        root = meta_doc.add_element("md:EntityDescriptor", namespaces)
+        root.attributes["ID"] = OneLogin::RubySaml::Utils.uuid
+        root.attributes["entityID"] = settings.sp_entity_id if settings.sp_entity_id
+        root.attributes["validUntil"] = valid_until.strftime('%Y-%m-%dT%H:%M:%S%z') if valid_until
+        root.attributes["cacheDuration"] = "PT" + cache_duration.to_s + "S" if cache_duration
+        root
+      end
+
+      def add_sp_sso_element(root, settings)
+        root.add_element "md:SPSSODescriptor", {
             "protocolSupportEnumeration" => "urn:oasis:names:tc:SAML:2.0:protocol",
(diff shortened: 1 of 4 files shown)


lib/onelogin/ruby-saml/metadata.rb
```diff
@@ -117,23 +142,33 @@ def generate(settings, pretty_print=false, valid_until=nil, cache_duration=nil)
...
+      def add_extras(root, _settings)
+        root
+      end
+
+      def embed_signature(meta_doc, settings)
+        return unless settings.security[:metadata_signed] && settings.private_key && settings.certificate
+
+        private_key = settings.get_sp_key
+        cert = settings.get_sp_cert
+        meta_doc.sign_document(private_key, cert, settings.security[:signature_method], settings.security[:digest_method])
```
pitbulk, line 158: meta_doc.sign_document(settings.get_sp_key, settings.get_sp_cert, settings.security[:signature_method], settings.security[:digest_method])

lib/xml_security.rb
```diff
@@ -161,10 +161,10 @@ def sign_document(private_key, certificate, signature_method = RSA_SHA1, digest_
       # add the signature
       issuer_element = self.elements["//saml:Issuer"]
       if issuer_element
-        self.root.insert_after issuer_element, signature_element
+        self.root.insert_after(issuer_element, signature_element)
       else
-        if sp_sso_descriptor = self.elements["/md:EntityDescriptor"]
-          self.root.insert_before sp_sso_descriptor, signature_element
+        if sp_sso_descriptor = self.elements["/md:EntityDescriptor/md:SPSSODescriptor"]
+          self.root.insert_before(sp_sso_descriptor, signature_element)
```
johnnyshields, line 167: @pitbulk the code as I have it here is correct with what you are saying. `self.root` is `EntityDescriptor`, so `self.root.insert_before(sp_sso_descriptor, ...)` means "Insert with `EntityDescriptor` as parent, as a sibling immediately before `md:SPSSODescriptor`)

lib/onelogin/ruby-saml/metadata.rb
johnnyshields: Done. Note the first line is `unless` not `if`, but otherwise exactly as you said.

lib/xml_security.rb
pitbulk: Ok, in case of md:EntitiesDescriptor Signature needs to be added as first child which is covered later... I think you are right.... maybe we could add a unittest to verify where the signature is added on the different scenarios to assure it is not break in the future. Rather than that the PR seems ok, I just had a minus comment about a refactor. Good job.

johnnyshields: In that case we should force it to always insert explicitly as the first child of `md:EntitiesDescriptor`. The way the code is now, it "happens to work" because `md:SPSSODescriptor` happens to be the first child. But if one's subclass were to do some crazypants custom modification of the xml in the `add_extras` method, this might no longer be the case.

johnnyshields: I'm not an expert in REXML so I'll have to check the right way to do this. If you know please let me know :)

lib/onelogin/ruby-saml/metadata.rb
pitbulk: you are right ;)

lib/xml_security.rb
pitbulk: Not an REXML expert neither ;)

johnnyshields: Unit test added and ready for final review. Note that unit test includes `validate_xml!` which catches most of errors. In a separate PR we should consider to have the Metadata class itself do the XSD validation, since we're now allowing users to add custom elements and the validation is quite strict.

johnnyshields:
@pitbulk good to merge?

pitbulk (approved):

pitbulk (approved):

PR #610, opened 2021-08-14
New IDP sso/slo binding params which deprecate :embed_sign

### LINKED ISSUE #606: :embed_sign vs. IdP SingleSignOnService Binding (HTTP Method)
Correct me if I'm wrong, but it appears the `:embed_sign` parameter is redundant with the IdP URL `Binding`. In other words:

- if the `Binding` is `urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST`, then you MUST have `embed_sign: true`
- else if the `Binding` is `urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect`, then you MUST have `embed_sign: false`

(Otherwise, your requests won't be signed, e.g. a `POST` with `embed_sign: false` will simply have no signature.)

Assuming this is the case, since `embed_sign` is a single value, it's not possible to mix SSO `POST` with SLO `Redirect`. Although not common in practice, it is theoretically possible that an IdP could require it that way in their metadata.

If this is the case, I wonder if there might be a better way to do this. For example, we could get rid of `:embed_sign` and introduce `Settings.idp_sso_binding` and `Settings.idp_slo_binding` parameters which would allow you to control which bindings you want to use for each endpoint.

johnnyshields:
Fixes #606

Adds settings.idp_sso_service_binding and `settings.idp_slo_service_binding`, and deprecates `settings.security[:embed_sign]`. If specified, new binding parameters will be used in place of `:embed_sign` to determine how to handle SAML message signing (`HTTP-POST` embeds signature and `HTTP-Redirect` does not.)

In addition:
- The `IDPMetadataParser` `parse`, `parse_to_hash` and `parse_to_array` methods now retrieve
`idp_sso_service_binding` and `idp_slo_service_binding`.
- For convenience you may now use the Symbol aliases `:post` and `:redirect` for any `settings.*_binding` parameter.
- Many of the getters/setters in the `Settings` class had needlessly verbose code. I've cleaned them up.


johnnyshields:
@pitbulk ready for your review

johnnyshields:
Conflicts resolved

pitbulk (approved):

johnnyshields:
@pitbulk really appreciate the prompt response! Thank you!

pitbulk:
Thank you for your recent contributions.

PR #619, opened 2021-09-11
Make SLO request signature verification compatible with ADFS

### LINKED ISSUE #470: Raw Parameters for SloLogoutRequest?
Any guidance on how to get the raw parameters in Ruby on Rails for the `raw_get_parameters` hash in the new SloLogoutResponse setup?

I ended up using `request.query_parameters`, which seems to skip the encoding processing that `params` goes through. Is that what you're referring to here or is there some other approach we should be looking at?

For reference to others, since I had a had time on this one, this is what we're using:

```
    raw_query_params = request.query_parameters
    options = {
      get_params: {
        "Signature" => params["Signature"]
      },
      raw_get_params: {
        "SAMLRequest" => raw_query_params["SAMLRequest"],
        "SigAlg" => raw_query_params["SigAlg"],
        "RelayState" => raw_query_params["RelayState"]
      },
      matches_request_id: request_id
    }
```

lorint:
(Fixes #470)

Why?
  The Microsoft Azure Active Directory or other ADFS related solutions have implemented
  SAML in a way that CGI-encoded request parameters such as SAMLRequest and SigAlg have
  hexidecimal values represented in lower-case.  This is not in keeping with how the
  vast majority of other SAML implementations operate, and as a result causes
  incompatibility with many environments.  In general it just seems that Microsoft
  likes to be different, and perhaps for dubious reasons.

How?
  In the OneLogin::RubySaml::SloLogoutrequest#validate_signature method, when creating
  raw_get_params and using them to calculate a sha256 hash used for verification,
  first attempt to retrieve any existing query string detail from the ACS URL, which
  will have preserved any non-standard casing which was provided by the IdP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@
 * Support Process Transform
 * Raise SettingError if invoking an action with no endpoint defined on the settings
 * Made IdpMetadataParser more extensible for subclasses
-*[#548](https://github.com/onelogin/ruby-saml/pull/548) Add :skip_audience option
+* [#548](https://github.com/onelogin/ruby-saml/pull/548) Add :skip_audience option
 * [#555](https://github.com/onelogin/ruby-saml/pull/555) Define 'soft' variable to prevent exception when doc cert is invalid
 * Improve documentation
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -9,6 +9,13 @@ to determine how to handle SAML message signing (`HTTP-POST` embeds signature an
 In addition, the `IdpMetadataParser#parse`, `#parse_to_hash` and `#parse_to_array` methods now retrieve
 `idp_sso_service_binding` and `idp_slo_service_binding`.
 
+Following on from the changes in version 1.6.0 to implement the `options[:raw_get_params]` parameter,
+when no raw params are specified, first attempt to build the querystring used for signature verification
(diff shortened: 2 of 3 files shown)


alagos:
I debuged my Azure SLO implementation for hours and I found that the problem was `ruby-saml` with this casing mismatch. I should've started looking into the existing issues/PR instead of spending so much time debugging. 
Thanks for the PR, I'll use your fork in the meantime that your changes aren't merged yet.

alagos:
A detail, this gem is supported [`for ruby >= 1.8.7`](https://github.com/onelogin/ruby-saml/blob/master/ruby-saml.gemspec#L23), but you're using `&.`, which is a 2.3 syntax.

pitbulk:
This is solved using properly the [lowercase_url_encoding](https://github.com/SAML-Toolkits/ruby-saml/blob/master/lib/onelogin/ruby-saml/settings.rb#L285) setting

At the [Single Log Out](https://github.com/SAML-Toolkits/ruby-saml#single-log-out) section of the README you can find it documented and used:

```
# Method to handle IdP initiated logouts
def idp_logout_request
  settings = Account.get_saml_settings
  # ADFS URL-Encodes SAML data as lowercase, and the toolkit by default uses
  # uppercase. Turn it True for ADFS compatibility on signature verification
  settings.security[:lowercase_url_encoding] = true
```

PR #627, opened 2021-11-11
Force escape downcasing for Azure SLO

alagos:
I followed the [Single Log Out](https://github.com/onelogin/ruby-saml#single-log-out) guide for my Azure AD integration and it worked all good until implementing that `idp_logout_request ` method where I received requests from Azure to log out a user. 
The requests come signed and trying to validate the signature it was failing, then I realised that the problem is that they are encoding the request parameters with downcase encoding characters (like using `%2f` instead of `%2F`) and they use the parameters downcased to generate the signature, therefore in validation time, when signed parameters are restored with `CGI.escape` they were different than the originals sent by Azure (all upcased).
To solve this problem, I added a `force_escape_downcasing` option for `OneLogin::RubySaml::SloLogoutrequest.new`, so all the signature verification is done with downcased encoded parameters.
I'm not sure where to add in the readme this option, specially because the SLO example I followed isn't verifying signed requests, but I'm open to suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ However, ruby-saml never enables this dangerous Nokogiri configuration;
 ruby-saml never enables DTDLOAD, and it never disables NONET.
 
 The OneLogin::RubySaml::IdpMetadataParser class does not validate in any way the URL
-that is introduced in order to be parsed. 
+that is introduced in order to be parsed.
 
 Usually the same administrator that handles the Service Provider also sets the URL to
 the IdP, which should be a trusted resource.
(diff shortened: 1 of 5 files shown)


pitbulk:
In php-saml we named this setting: [lowercaseUrlencoding](https://github.com/onelogin/php-saml/blob/master/advanced_settings_example.php#L121)

Can you:
- rename it and make it available at settings level instead be a options of the constructor?
- Apply it as well to [LogoutResponse](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/logoutresponse.rb) validations

alagos:
@pitbulk about your request:
> Rename it and make it available at settings level instead be a options of the constructor?

added as `settings.security[:lowercase_url_encoding]`, so done.

> Apply it as well to LogoutResponse validations

I guess you meant to change [the opposite method where the signature is unescaped](https://github.com/onelogin/ruby-saml/blob/master/lib/onelogin/ruby-saml/logoutresponse.rb#L218), right?
I don't think that will be really necessary, as `CGI.unescape` doesn't make any difference about the received string casing, so it will return always the same result:

```ruby
irb(main):008:0> CGI.unescape('%c3%a1sdf')
=> "ásdf"
irb(main):009:0> CGI.unescape('%C3%A1sdf')
=> "ásdf"
```

pitbulk (approved):

PR #629, opened 2022-01-06
Get tests running under Ruby 3.1 and add to CI

petergoldstein:
As Ruby 3.1 is out it seems like a good idea to get CI running with that version.

There was an issue with Timecop and argument handling that prevented tests from running successfully.  This was due to the gemspec pinning to a very old version of Timecop.  Updating that pin allowed Ruby 3.1 specs to pass without causing any additional spec failures.


petergoldstein:
@pitbulk Sure.  Let me know if the change I just pushed works.

pitbulk (approved):

PR #630, opened 2022-01-14
Allow optional prefix & suffix for AuthNRequest ID attribute

gowda:
`ID` attribute is returned as `InResponseTo` attribute of the response by servers (as mentioned by [Azure documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/single-sign-on-saml-protocol#authnrequest)).


pitbulk:
It makes more sense to add the prefix to the [uuid](https://github.com/PranaDiabetes/ruby-saml/blob/bdccdc2c3bacead2e386b77b75a40cfe28fbb0ac/lib/onelogin/ruby-saml/utils.rb#L254) method ot the Utils, rather than in only the request constructor.

You can add a constant to the Utils class by default "_" so then later the user will be able to override it dinamycally

gowda:
> It makes more sense to add the prefix to the [uuid](https://github.com/PranaDiabetes/ruby-saml/blob/bdccdc2c3bacead2e386b77b75a40cfe28fbb0ac/lib/onelogin/ruby-saml/utils.rb#L254) method ot the Utils, rather than in only the request constructor.
> 
> You can add a constant to the Utils class by default "_" (the current prefix) so then later the user will be able to override it dinamycally

I have update `Utils.uuid` method to accept an optional `prefix` which is set to `_` by default.

pitbulk:
I dislike the need of changing code in all the places where uuid is used with the new parameter.

Can you instead follow my suggestion and just add a param to the Util class, by default "_" and then change the uuid in order to use it.

developers that want the new behavior only will need to change the attribute from the Util class and that all.

pitbulk:
prefix support gonna be included at #633

PR #647, opened 2022-12-14
replace "strip!" from SAML signing digest

bramleyjl:
## Description
This PR is for a fix to the `compute_digest` method that was rendering a `nil` digest after Base64-encoding it. The reason for this bug is the usage of [strip!](https://apidock.com/ruby/v2_5_5/String/strip%21), a method that returns `nil` if no whitespace is found on either end of the string to be stripped. By replacing it with [strip](https://apidock.com/ruby/v2_5_5/String/strip), an identical method that will return the original string if no whitespace is found to be stripped, this bug can be prevented.

We on the VA.gov development team have been using a [forked version of the Ruby-SAML gem](https://github.com/department-of-veterans-affairs/ruby-saml) with this change since October and have not had any issue with it.

## External Links
Link to original issue: https://github.com/SAML-Toolkits/ruby-saml/issues/643
Link to code at issue: https://github.com/SAML-Toolkits/ruby-saml/blob/master/lib/xml_security.rb#L178


pitbulk:
Fixed in #650

Issue #603, opened 2021-08-08
IdpMetadata#name_id_format(s) should be parsed as a prioritized list


johnnyshields:
Currently `IdpMetadata#name_id_format` is a singular (scalar) value.

Since the IdP metadata can include multiple nameId formats, it would be useful to add a #name_id_formats method.

Then, `#name_id_format` could be selected to be the one that matches the SP metadata requirement.

johnnyshields:
I think the right way would be to add a new option to the `IdpMetadata#parse_method` which mirrors how the slo/sso bindings are done:

```
# @option options [Array<String>, nil] :name_id_format an ordered list of NameIDFormats to detect a desired value. The first NameIDFormat in the list that is included in the metadata will be used.
```

(Note that we currently have the following:)

```
# @option options [Array<String>, nil] :sso_binding an ordered list of bindings to detect the single signon URL. The first binding in the list that is included in the metadata will be used.
# @option options [Array<String>, nil] :slo_binding an ordered list of bindings to detect the single logout URL. The first binding in the list that is included in the metadata will be used.
# @option options [String, nil] :entity_id when this is given, the entity descriptor for this ID is used. When ommitted, the first entity descriptor is used.
```

Would you accept a PR for this?

pitbulk:
Seems reasonable

johnnyshields:
PR raised

PR #650, opened 2023-01-02
Replace strip! by strip on compute_digest method

pitbulk:
This PR replaces #647, Issue reported at #603

@bramleyjl This PR contains only 1 commit vs the one you created with 6 commits.

This PR is for a fix to the compute_digest method that was rendering a nil digest after Base64-encoding it. The reason for this bug is the usage of [strip!](https://apidock.com/ruby/v2_5_5/String/strip%21), a method that returns nil if no whitespace is found on either end of the string to be [strip](https://apidock.com/ruby/v2_5_5/String/strip)ped. By replacing it with strip, an identical method that will return the original string if no whitespace is found to be stripped, this bug can be prevented.

### DIFF
--- a/lib/xml_security.rb
+++ b/lib/xml_security.rb
@@ -177,7 +177,7 @@ module XMLSecurity
 
     def compute_digest(document, digest_algorithm)
       digest = digest_algorithm.digest(document)
-      Base64.encode64(digest).strip!
+      Base64.encode64(digest).strip
     end
 
   end


RDeckard:
Indeed, as lots of Ruby bang methods, `#strip!` returns `nil` if it didn't make any change to its receiver. 🎯

But those methods are also known to be more efficient (especially with big objects), as they mutate their receivers instead of duplicate it.

So if you want the best of both worlds (knowing that here the mutation didn't generate any problem by itself), you could also do `receiver.tap(&:strip!)`.
You probably already know that, but the short explanation is that the `#tap` method allows us to do whatever we want with its receiver in its block, then returns _the receiver itself_ (i.e. _not the result of the block_), therefore in that case it will never returns `nil`. 🙂

Nevertheless I am not familiar with this gem, so I'm not sure at all that it could be any significant impact in this case.

PR #656, opened 2023-03-08
Add Rubocop + some basic fixes

johnnyshields:
This PR adds Rubocop, then fixes a few basic things like:
- Use frozen strings everywhere
- Convert hash => rockets to use key: value syntax
- Normalize if/unless

I plan to do a lot more Rubocop fixes, but as there will be a lot of code changing probably better to do them in a series of smaller PRs.


johnnyshields:
@pitbulk this is ready for review. You can ignore the Coveralls failures (-0.06%)

pitbulk:
Changes seems ok 👌

johnnyshields:
@pitbulk I don't think there is any reason to maintain master branch anymore, aside from super-critical security fixes.

Suggest to make current master to be 1.x-stable, and make 2.x the new master.

johnnyshields:
> Some devs could be using master on their end, and I dont want to break their apps.

It's certainly fine to put out an announcement, but the possibility of "breakage" here is pretty remote. All of the following would need to be true for a breakage to occur:
1. Their Gemfile is pointing to `master` without a `ref:` specified
2. They are using Ruby 2.5 or earlier -- a deep EOL version (Ruby 2.6+ won't break)
3. They are doing something insane like `bundle update` on every install, or deleting `Gemfile.lock`, since normal `bundle install` will lock to the commit ref.

If the developer is doing all of the above 3 things wrong, we are doing the developer a favor by causing their app to break, so they can learn about best practices.

PR #660, opened 2023-03-09
Remove Onelogin namespace

### LINKED ISSUE #659: Remove OneLogin namespace (in a major version upgrade)
I propose that we remove the `OneLogin` root namespace in a major version upgrade, perhaps we do it in 2.0. The root namespace should be `RubySaml`. It seems OneLogin is no longer the sponsor of the project.

johnnyshields:
Fixes #659

Note this moves file paths from `onelogin/ruby-saml` to `ruby_saml`. The gem still loads with `require 'ruby-saml'` (hyphenated)

I will update README to remove mentions of Onelogin in a future PR.


pitbulk:
@johnnyshields removing the namespace will break a lot of current projects that use this gem as dependency.
I'm not confortable with such change.

johnnyshields:
> In order to keep compatiblity, what do you think if we git mv all the files from onelogin/ruby-saml to ruby_saml folder, then apply the changes, and then we createa new empty onelogin/ruby-saml that will point to the ruby-saml.
Or maybe there is a cleaner way to do that.

Sure, that sounds good.

pitbulk:
@johnnyshields  I guess is better to reapply this change with the new merged files from v2

johnnyshields:
I'll update this PR soon!

johnnyshields:
Replaced by https://github.com/SAML-Toolkits/ruby-saml/pull/685

PR #664, opened 2023-06-12
Forward soft argument to validate_signature

neanias:
When calling the `validate_document` method chain, the `soft` parameter is passed down from one method to another except in the hand off from `validate_document_with_cert` to `validate_signature`.

### DIFF
--- a/lib/xml_security.rb
+++ b/lib/xml_security.rb
@@ -262,7 +262,7 @@ module XMLSecurity
       else
         base64_cert = Base64.encode64(idp_cert.to_pem)
       end
-      validate_signature(base64_cert, true)
+      validate_signature(base64_cert, soft)
     end
 
     def validate_signature(base64_cert, soft = true)


neanias:
For 1. I would think it was `true` in the first place because validation may have always wanted to be soft at this stage to avoid raising an error? It's a bit unusual to me in reading the git log that when the `soft` parameter was introduced for this method it was to allow surpressing exceptions, but there's no way to _stop_ surpressing exceptions. I'm not clear if the tests are covering the unhappy path as a result since it doesn't check if passing `soft = false` raises an error.

For 2. it's been over a year since I opened this so I can't exactly remember _why_ I opened the PR in the first place. I think I was trying to understand why my code wasn't throwing any errors when I passed in `soft = false` as per the docs.

neanias:
The `soft = true` param was added in this commit: https://github.com/neanias/ruby-saml/commit/d0e117a5d9a391c668793cd7a14ae010699a73ef

johnnyshields:
OK. @pitbulk I think it's reasonable to merge this.

pitbulk:
I remember I checked it time ago when was registered, but for some reason I did not merge it.

As "validate signature" is a critical piece, I will need more time to review this PR and check why it was implemented in that way.

johnnyshields:
This change seems important for security.

The previous logic made signature validation _always_ a soft validation. However, if one's system is relying on errors to be raised, then currently failed signature validation **won't** raise an error, and one will permit a failed signature validation when one should actually reject it! 😱

I can't image a case where a user wants _hard_ errors, but only wants signature validation to be _soft_.

neanias:
Also realised that I didn't add tests to check that the right _hard_ error is raised when soft is disabled. I'll add one now

pitbulk:
If you check the use of validate_document_with_cert in the [code](https://github.com/SAML-Toolkits/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L885), it is already used with soft parameter with true value, and the error is [added](https://github.com/SAML-Toolkits/ruby-saml/blob/master/lib/onelogin/ruby-saml/response.rb#L906) as in the rest of validations.

neanias:
I see. So does this method need to accept the `soft` flag at all then?

PR #670, opened 2023-09-26
Correct reference to issuer / idp_entity_id (was sp_entity_id)

Capncavedan:
I believe [this note](https://github.com/SAML-Toolkits/ruby-saml/pull/670) referencing `sp_entity_id` in the upgrade guide is incorrect, and should reference `idp_entity_id` instead.

> deprecates the use of `settings.issuer` in favour of `settings.sp_entity_id`

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -24,7 +24,7 @@ in favor of `idp_sso_service_url` and `idp_slo_service_url`. The `IdpMetadataPar
 
 ## Upgrading from 1.10.x to 1.11.0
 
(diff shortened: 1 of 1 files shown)


pitbulk:
@Capncavedan the note is correct.

See commit:
https://github.com/SAML-Toolkits/ruby-saml/commit/660fc1071b59033d199ec9df6e0f8791b56c876b

The issuer is who generates the SAML Message, in this case is the SP, so the SAML Messages generated by the SP has as Issuer the SP Entity ID.

Capncavedan:
Thank you for the explanation, @pitbulk

PR #677, opened 2024-02-28
Allow passing in timeout/retry settings to Net::HTTP

tjschuck:
When fetching remote XML files from arbitrary URLs, you might want to configure different values for timeouts/retries to avoid allowing users to DoS you via intentionally slow endpoints.  The Net::HTTP defaults are 60 seconds plus 1 retry, which could easily deplete resources if intentionally exploited.

### DIFF
--- a/lib/onelogin/ruby-saml/idp_metadata_parser.rb
+++ b/lib/onelogin/ruby-saml/idp_metadata_parser.rb
@@ -54,6 +54,11 @@ module OneLogin
       # @param url [String] Url where the XML of the Identity Provider Metadata is published.
       # @param validate_cert [Boolean] If true and the URL is HTTPs, the cert of the domain is checked.
       #
+      # @param request_options [Hash] options used for requesting the remote URL
+      # @option request_options [Numeric, nil] :open_timeout Number of seconds to wait for the connection to open. See Net::HTTP#open_timeout for more info. Default is the Net::HTTP default.
+      # @option request_options [Numeric, nil] :read_timeout Number of seconds to wait for one block to be read. See Net::HTTP#read_timeout for more info. Default is the Net::HTTP default.
+      # @option request_options [Integer, nil] :max_retries Maximum number of times to retry the request on certain errors. See Net::HTTP#max_retries= for more info. Default is the Net::HTTP default.
+      #
       # @param options [Hash] options used for parsing the metadata and the returned Settings instance
       # @option options [OneLogin::RubySaml::Settings, Hash] :settings the OneLogin::RubySaml::Settings object which gets the parsed metadata merged into or an hash for Settings overrides.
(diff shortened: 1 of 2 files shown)


johnnyshields:
@pitbulk this looks reasonable to merge.

pitbulk:
@johnnyshields  I think its ok, but I gonna prefer ``request_options`` to be the last paramter of ``parse_remote`` and ``parse_remote_to_array`` methods, rather than include them before the ``options`` parameter.

@tjschuck can you do the change?

tjschuck:
@pitbulk Pushed!

Thanks @pitbulk and @johnnyshields

johnnyshields:
Sorry to bikeshed this, I misread @pitbulk's earlier comment. I was thinking we would make these just part of `options`. But I guess it's fine this way too.

tjschuck:
@johnnyshields That's fine with me too.  I honestly can't remember why I did it this way...

If you want me to push it into `options`, just let me know and I can do that shortly.

pitbulk:
@johnnyshields I think it's ok to have both options differentiated. As we are introducing now all related to the request, I'm not against to leave its own place. Wonder if there is a way to inject them in the request rather than initialize each parameter one by one.
```
 http.open_timeout = request_options[:open_timeout] if request_options[:open_timeout]
 http.read_timeout = request_options[:read_timeout] if request_options[:read_timeout]
 http.max_retries = request_options[:max_retries] if request_options[:max_retries]
```

lib/onelogin/ruby-saml/idp_metadata_parser.rb
```diff
@@ -80,11 +85,16 @@ def parse_remote(url, validate_cert = true, options = {})
...
+      # @option request_options [Numeric, nil] :read_timeout Number of seconds to wait for one block to be read. See Net::HTTP#read_timeout for more info. Default is the Net::HTTP default.
+      # @option request_options [Integer, nil] :max_retries Maximum number of times to retry the request on certain errors. See Net::HTTP#max_retries= for more info. Default is the Net::HTTP default.
+      #
       # @return [Hash]
       #
       # @raise [HttpError] Failure to fetch remote IdP metadata
-      def parse_remote_to_hash(url, validate_cert = true, options = {})
-        parse_remote_to_array(url, validate_cert, options)[0]
+      def parse_remote_to_hash(url, validate_cert = true, options = {}, request_options = {})
+        parse_remote_to_array(url, validate_cert, request_options, options)[0]
```
johnnyshields, line 97: oops this is wrong position

johnnyshields:
Using one `options` arg avoids human errors like the one I just caught. I strongly prefer it.

pitbulk:
Also this should go on branch V2.0.0 where we force new ruby version

johnnyshields:
@tjschuck I'll take this one over.

tjschuck:
@johnnyshields Go nuts!  Thanks.

johnnyshields:
Replaced by #709

PR #679, opened 2024-03-28
Make the gem compatible with `--enable-frozen-string-literal`

casperisfine:
This option has been available for a very long time in Ruby, and will probably become the default in the semi-near future.

Ref: https://bugs.ruby-lang.org/issues/20205

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, macos-latest]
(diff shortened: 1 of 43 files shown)


johnnyshields:
@casperisfine it's better to just add the magic comment to the top of every file. I think Rubocop autocorrect can do it.

casperisfine:
@johnnyshields I can do that if you prefer, that's no problem.

johnnyshields:
@pitbulk are you ok with it?

casperisfine:
73b6683c2a202331e74b9890a42d7809a8f7b40a

casperisfine:
Hum, looks like `String#+@` was only added in `2.3`:

```
undefined method `+@' for "_":String (NoMethodError)
```

So I may need to use `String#dup` to preserve compatibility.

casperisfine:
> Also should be a single blank line after the magic comment.

Done.

johnnyshields:
This is already fixed in v2.x branch, so this PR can be closed. @pitbulk

PR #683, opened 2024-07-07
[WIP - NOT READY] Support DSA and ECDSA signing keys

johnnyshields:
Fixes https://github.com/SAML-Toolkits/ruby-saml/issues/661

Currently RubySaml supports only RSA keys. The SAML standard can also support ECDSA and DSA keys. This PR adds support for both:
1. Validating IdP EC/DSA sigs **AND**
2. Using SP EC/DSA signing keys.

It includes the following changes, which are all done in a backward compatible manner:
- When generating SP metadata/requests, `settings.security[:signature_method]` now ignores the "rsa" component of its user-set value and automatically uses whatever type of SP signing public key you actually set (e.g. a DSA key) plus the "sha" component of the value.
   - (Previously, only RSA was supported, so this doesn't break anything.)
- `settings.security[:signature_method]` supports shortcut values `:sha1`, `:sha256`, etc.
   - Shortcuts `:rsa_sha256`, `:dsa_sha256` etc. also work, but as per above the "rsa"/"dsa" are ignored in favor of the SP public key type.
- Similar to above `settings.security[:digest_method]` supports shortcut values `sha1`, `sha256`, etc.
- New module `XMLSecurity::Crypto` is extracted from `XMLSecurity::Document`
- Cleaned-up code, including related to canonicalization

Things this PR does **NOT** do:
- I haven't yet looked at encryption. Probably, DSA/EC keys won't work with encryption (I'm unsure whether they actually _should_ work, given that DSA is supposedly for signing-only.) To use the functionality in this PR, you must either:
   - Disable SP encryption when using EC/DSA keys **OR**
   - Use separate SP signing vs. encryption keys (supported in #673), and make sure the encryption key is an RSA key.

**TODO:**
- [x] Ensure existing tests pass
- [ ] Test SP EC keys on metadata
- [ ] Test SP EC keys on messages
- [ ] Test SP DSA keys on metadata
- [ ] Test SP DSA keys on messages
- [ ] Test IdP EC keys on metadata
- [ ] Test IdP EC keys on messages
- [ ] Test IdP DSA keys on metadata
- [ ] Test IdP DSA keys on messages
- [ ] Investigate encryption

### DIFF
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -75,14 +75,14 @@ module OneLogin
         sp_signing_key = settings.get_sp_signing_key
 
         if settings.idp_sso_service_binding == Utils::BINDINGS[:redirect] && settings.security[:authn_requests_signed] && sp_signing_key
-          params['SigAlg'] = settings.security[:signature_method]
+          params['SigAlg'] = settings.get_sp_signature_method
           url_string = OneLogin::RubySaml::Utils.build_query(
             :type => 'SAMLRequest',
             :data => base64_request,
             :relay_state => relay_state,
             :sig_alg => params['SigAlg']
           )
-          sign_algorithm = XMLSecurity::BaseDocument.new.algorithm(settings.security[:signature_method])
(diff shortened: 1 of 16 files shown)


johnnyshields:
I'll rebase this on v2.0 branch once other PRs are merged.

PR #684, opened 2024-07-08
v2.0: Merge latest master into v2.0

johnnyshields:
This PR catches up v2.x branch with the latest master code.

This PR also drops support for Ruby 2.6, 2.7 and JRuby 9.3 all of which are now both EOL. (Even Ruby 3.0 is EOL, but we can support it for now.)


johnnyshields:
@pitbulk this is ready for your review now.

lib/onelogin/ruby-saml/version.rb
```diff
@@ -2,6 +2,6 @@
 
 module OneLogin
   module RubySaml
-    VERSION = '1.15.0'
+    VERSION = '1.16.0'
```
pitbulk, line 5: @johnnyshields  I'm more inclined to move to a 2.X version now that we have dropped support for old Ruby versions.

That way we can use 1.16 and 1.17 for security bugfixes and key features we want to add to the old branch.
We should also update the changelog properly.

pitbulk:
I only have the concern of the ruby-saml version, the rest seems good to be merged.

lib/onelogin/ruby-saml/version.rb
johnnyshields: Done! Changed to 2.0.0

PR #685, opened 2024-07-08
v2.0: Remove OneLoginNamespace

johnnyshields:
Fixes https://github.com/SAML-Toolkits/ruby-saml/issues/659

Note this moves file paths from `lib/onelogin/ruby-saml` to `lib/ruby_saml`. The gem still loads with `require 'ruby-saml'` (hyphenated) as before.

By the way, I have used the `git mv` command in this PR. You can see it specifically in this commit: https://github.com/SAML-Toolkits/ruby-saml/pull/685/commits/7c90d25160502920465ccdd1524f22a2e919924c. The [following commit](https://github.com/SAML-Toolkits/ruby-saml/pull/685/commits/d1d3aeb53af0f06eb329929d20891cde05639f7c) is purely modification of those files. When you view the full diff of the PR, Github's UI makes it looks like certain files were deleted and re-created, but this is not actually the case in Git itself.


johnnyshields:
@pitbulk this is ready for your review and merge.

pitbulk:
@johnnyshields Can you fix the conflict with version?
Why there are all those commits? Maybe makes sense if you sync your project with v2.x branch and just move the mv and push?

Also, we need to ensure backward compatibility. Can you add some tests using the OneLogin domain: 
```
OneLogin::RubySaml
```

Thanks for your hard work!

PR #689, opened 2024-07-08
v2.0: Remove settings.compress_request and compess_response parameters

johnnyshields:
Fixes https://github.com/SAML-Toolkits/ruby-saml/issues/676

Remove the `settings.compress_request` and `settings.compress_response` parameters. Set their behavior automatically based on `settings.idp_sso_service_binding` and `settings.idp_slo_service_binding` respectively. `HTTP-Redirect` will always use compression, while `HTTP-POST` will not.

(Compressing HTTP-POST is non-sensical, while most SAML services will fail if you don't compress redirects, and you may even hit the 2048 char URL limit.)


pitbulk:
Im worried about other libraries using settings.compress_request and settings.compress_response for building its own bindings, not covered by RubySaml.

Can we instead of removing the settings, give the "default" binding behavior if no settings.compress_request and settings.compress_response was set?

johnnyshields:
@pitbulk how about we keep the parameters but remove their functionality and simply have them display a "deprecated" message?

The problem with these parameters is that POST binding should never be compressed, and Redirect binding should always be compressed. Any other configuration is invalid, so continuing to support these will only cause issues.

johnnyshields:
(Another option is to do the above deprecation on 1.7.0, and remove entirely in 2.0.0)

pitbulk:
My proposal:

At settings.rb, [set the default](https://github.com/SAML-Toolkits/ruby-saml/blob/master/lib/onelogin/ruby-saml/settings.rb#L280) values as:
```
:compress_request => nil,
:compress_response => nil,
```

Create at utils.rb an auxiliary like:
```
     # Calculates if compression is required
     # @param type [String] SAMLRequest | SAMLResponse
     # @param service [String] sso | slo
     # @param settings [OneLogin::RubySaml::Settings|nil] Toolkit settings 
    def may_compress_message(type, service, settings):
      if settings.compress_request.present? and type == "SAMLRequest"  # Old behavior
        compress = settings.compress_request
      else if settings.compress_response.present? and type == "SAMLResponse"  # Old behavior
        compress = settings.compress_response
      else if service == "sso"
         compress = is_binding_redirect(settings.idp_sso_service_binding)
      else if service == "slo"
          compress = is_binding_redirect(settings.idp_slo_service_binding)
      else 
          compress = false
       return compress

    def is_binding_redirect(settings_binding):
      return settings_binding == Utils::BINDINGS[:redirect]
```

Then in authrequest.rb for example:
```
  request = deflate(request) if Utils::may_compress_message("SAMLRequest" , "sso", settings)
  
  ...
  
  if Utils::is_binding_redirect(settings.idp_sso_service_binding) && settings.security[:authn_requests_signed] && sp_signing_key
```

Or at slo_logoutresponse.rb
```
response = deflate(response) if Utils::may_compress_message("SAMLResponse" , "slo", settings)
  
  ...
  
  if Utils::is_binding_redirect(settings.idp_slo_service_binding) && settings.security[:logout_responses_signed] && sp_signing_key

```


Also, Leave but add a deprecation warning to:
```
      # Deflate, base64 encode and url-encode a SAML Message (To be used in the HTTP-redirect binding)
      # @param saml [String] The plain SAML Message
      # @param settings [OneLogin::RubySaml::Settings|nil] Toolkit settings
      # @return [String] The deflated and encoded SAML Message (encoded if the compression is requested)
      #
      def encode_raw_saml(saml, settings)
        saml = deflate(saml) if settings.compress_request

        CGI.escape(encode(saml))
      end
```
And create a new 
```
      # Deflate (or not), base64 encode and url-encode a SAML Message (To be used in the HTTP-redirect binding)
      # @param saml [String] The plain SAML Message
      # @param compress [true|false] Whether or not the SAML should be deflated.
      # @return [String] The SAML Message (deflated if requested) base64 encode and url-encode
      #
      def encode_raw_saml_message(saml, compress)
        saml = deflate(saml) if compress

        CGI.escape(encode(saml))
      end
```

johnnyshields:
@pitbulk Is your concern is *specifically* about the `encode_raw_saml` method?

How about this:

1. Leave in `Settings.compress_request`, but make it do nothing.
2. 

```ruby
def encode_raw_saml_message(saml, settings_or_compress)
  saml = deflate(saml) if settings_or_compress.is_a?(TrueClass)

  unless settings_or_compress.compress_request.nil?
    log_deprecation_warning("[WARNING] Please change the second argument of `encode_raw_saml_message` to a boolean indicating whether or not to use compression. Using a boolean will be required in RubySaml 2.1.0")
    saml = deflate(saml) if settings_or_compress.compress_request
  end

  CGI.escape(encode(saml))
end
```
3. Remove `Settings.compress_response` as it does not affect `encode_raw_saml_message` at all.

pitbulk:
My concern is that people could be using Settings.compress_request and Settings.compress_response at its code.
We can mark it as deprecated in 2.0 and maybe remove in future 2.1
Are you against the use of the aux may_compress_message?

I'm ok with the encode_raw_saml idea you proposed, but maybe:

```
def encode_raw_saml(saml, settings_or_compress)
  if settings_or_compress.is_a?(TrueClass)
    saml = deflate(saml) if settings_or_compress
  else
    log_deprecation_warning("[DEPRECATION WARNING] Please change the second argument of `encode_raw_saml_message` to a boolean indicating whether or not to use compression. Using a boolean will be required in RubySaml 2.1.0.")

    if settings_or_compress.respond_to?(:compress_request)
      saml = deflate(saml) if settings_or_compress.compress_request
    end
  end
  CGI.escape(encode(saml))
end
```

And later in 2.1, we simply:
```
def encode_raw_saml(compress)
  saml = deflate(saml) if compress
  CGI.escape(encode(saml))
end
```

johnnyshields:
> Are you against using the aux may_compress_message
Yes, I really think it's not needed. We should simply inform people to use the existing `idp_sso/slo_service_binding` params, rather than making a new `may_compress_message` that requires them to think about legacy params. Nothing will break regardless.

> I'm ok with the encode_raw_saml idea you proposed, but maybe
Your code raises a dep message if `settings_or_compress` is false or nil. Correcting my proposed code here:

```ruby
def encode_raw_saml(saml, settings_or_compress)
  saml = deflate(saml) if settings_or_compress.is_a?(TrueClass)

  if settings_or_compress.respond_to?(:compress_request)
    log_deprecation_warning("[DEPRECATION WARNING] Please change the second argument of `encode_raw_saml_message` to a boolean indicating whether or not to use compression. Using a boolean will be required in RubySaml 2.1.0.")
    saml = deflate(saml) if settings_or_compress.compress_request
  end

  CGI.escape(encode(saml))
end
```

> And later in 2.1
Yes agreed to this

johnnyshields:
Moved to #695

PR #691, opened 2024-07-08
Make IdpMetadataParser#get_idp_metadata public

calebhearth:
I have a use-case for being able to cache the intermediate fetch of
metadata in case of temporary failures, so rather than:

    parser.parse_remote(url)

I'd like to

    begin
      metadata = parser.get_idp_metadata(url, true)
      do_my_caching(metadata)
      parser.parse(metadata)
    rescue HttpError
      load_cache
    end

There's a fair amount of logic in the `get_idp_metadata` method that I'd
rather not need to re-implement. Right now I have this implemented with
`parser.send(:get_idp_metadata, url, true)` which is obviously not great
if the internals of this class change in the future. Can we move this
method to the public API?


johnnyshields:
Seems reasonable. @pitbulk I vote to merge.

PR #692, opened 2024-07-09
v2.0: Move XMLSecurity namespace to RubySaml::XML

johnnyshields:
This change is done in the same manner as the OneLogin removal. It also splits out the multiple inner classes into individual files.

The PR also includes a test for alias XMLSecurity = RubySaml::XML


pitbulk:
@johnnyshields 
What happens if people in its code has reference for:
```
require 'lib/xml_security.rb'
```

Can we keep for backward compatiblity by leaving the xml_security.rb that gonna include the requirements of the new files, as we did with ruby-saml.rb?

johnnyshields:
@pitbulk changes done as requested.

PR #695, opened 2024-07-09
v2.0: Deprecate compress_request and compress_response

johnnyshields:
Fixes https://github.com/SAML-Toolkits/ruby-saml/issues/676

Replaces PR #689 

Deprecate `compress_request` and `compress_response` parameters. Set their behavior automatically based on settings.idp_sso_service_binding and settings.idp_slo_service_binding respectively. HTTP-Redirect will always use compression, while HTTP-POST will not.

As discussed the `encode_raw_saml` method has been made backward compatible.

This also contains light refactoring of `RubySaml::Logging` to add a `deprecate` method and be a singleton module without changing any of its existing method interfaces.


test/logoutrequest_test.rb
```diff
@@ -152,21 +152,7 @@ class RequestTest < Minitest::Test
         assert_match %r[<ds:DigestMethod Algorithm='http://www.w3.org/2001/04/xmlenc#sha256'/>], inflated
       end
 
-      it "create a signed logout request" do
-        settings.compress_request = true
```
pitbulk, line 156: @johnnyshields 
Should we instead remove only the:
```
settings.compress_request = true
```
and use
```
settings.idp_slo_service_binding = RubySaml::Utils::BINDINGS[:post]
```
So we keep the test?

pitbulk:
Added a minor comment, rest seems ok to be merged.

test/logoutrequest_test.rb
johnnyshields: You have very good eyes! 👀

It was removed because when you remove compression from this case (as required for HTTP-POST) and test the `create_params` method (`create` is not used for POST), then it becomes identical to the test immediately below it.

lib/ruby_saml/settings.rb
```diff
@@ -296,8 +292,40 @@ def get_binding(value)
       }.freeze
     }.freeze
 
+    # @deprecated Will be removed in v2.1.0
+    def compress_request
+      compress_deprecation('compress_request', 'idp_sso_service_binding')
+      defined?(@compress_request) ? @compress_request : true
```
johnnyshields, line 298: One more thing I caught, the previous default for these was "true", so it needs to stay that way.

PR #699, opened 2024-07-09
v2.0: Settings params deprecations

johnnyshields:
Deprecate `RubySaml#Settings` params and alias them to the new ones. All these deprecations were declared in CHANGELOG in earlier versions already. *This does not affect any existing behavior.*

- issuer --> sp_entity_id
- idp_sso_target_url --> idp_sso_service_url
- idp_slo_target_url --> idp_slo_service_url
- assertion_consumer_logout_service_url --> single_logout_service_url
- assertion_consumer_logout_service_binding --> single_logout_service_binding

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [#686](https://github.com/SAML-Toolkits/ruby-saml/pull/686) Use SHA-256 as the default hashing algorithm everywhere instead of SHA-1, including signatures, fingerprints, and digests.
 * [#695](https://github.com/SAML-Toolkits/ruby-saml/pull/695) Deprecate `settings.compress_request` and `settings.compess_response` parameters.
 * [#690](https://github.com/SAML-Toolkits/ruby-saml/pull/690) Remove deprecated `settings.security[:embed_sign]` parameter.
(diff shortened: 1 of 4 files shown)


johnnyshields:
@pitbulk ready for review and merge.

pitbulk:
Merging this one, but we need to be careful when we move to 2.1

Maintain the code clean is a nice to have, but if during doing this you break several projects is not cool.

If you execute a simple [search on Github](https://github.com/search?q=%22settings.issuer%22+language%3ARuby+&type=code), you will find that a tons of projects still use ``settings.issuer``, so imagine all those projects using ruby-saml, that added SAML support years ago (when this parameter was in use). I'm pretty sure that is really possible that the developer that made the integration could not be anymore in the company, and nobody gonna be aware of the change that is required, or will not notice of the need of the change until the integration broke.

johnnyshields:
I'm well aware. I've maintained an app for 10 years. I'm taking a lot of care not to break anything.

Even in version 2.1.0, I am thinking we should keep the methods but raise a hard RuntimeError `Settings#issuer parameter has been removed in RubySaml v2.1.0. Please downgrade to RubySaml v2.0.0 and fix all deprecation warnings first before upgrading to v2.1.0.`

PR #705, opened 2024-07-10
[READY] v2.0: Support DSA and ECDSA signing keys

johnnyshields:
Fixes https://github.com/SAML-Toolkits/ruby-saml/issues/661

Replaces https://github.com/SAML-Toolkits/ruby-saml/pull/683

Depends on https://github.com/SAML-Toolkits/ruby-saml/pull/711, please merge it first.

Currently RubySaml supports only RSA keys. The SAML standard can also support ECDSA and DSA keys. This PR adds support for both:
1. Validating IdP EC/DSA sigs **AND**
2. Using SP EC/DSA signing keys.

JRuby supports DSA but cannot support ECDSA due to [this issue](https://github.com/jruby/jruby-openssl/issues/257).

You may not use EC/DSA keys for encryption. An `ArgumentError` is raised if you attempt to do so.

It includes the following changes, which are all done in a backward compatible manner:
- When generating SP metadata/requests, `settings.security[:signature_method]` now ignores the "rsa" component of its user-set value and automatically uses whatever type of SP signing public key you actually set (e.g. a DSA key) plus the "sha" component of the value.
   - (Previously, only RSA was supported, so this doesn't break anything.)
- `settings.security[:signature_method]` supports shortcut values `:sha1`, `:sha256`, etc.
   - Shortcuts `:rsa_sha256`, `:dsa_sha256` etc. also work, but as per above the "rsa"/"dsa" are ignored in favor of the SP public key type.
- Similar to above `settings.security[:digest_method]` supports shortcut values `sha1`, `sha256`, etc.
- New module `XMLSecurity::Crypto` is extracted from `XMLSecurity::Document`
- Cleaned-up code, including related to canonicalization

**DONE:**
- [x] Ensure existing tests pass
- [x] request_test.rb -- cover all 3 key types
- [x] logoutrequest_test.rb -- cover all 3 key types
- [x] slo_logoutresponse_test.rb -- cover all 3 key types
- [x] metadata_test.rb -- cover all 3 key types
- [x] Fix JRuby failures
- [x] basic tests (settings, etc.) -- cover all 3 key types
- [x] test EC/DSA keys when used by IdP
- [x] disable encryption with EC/DSA keys


johnnyshields:
@pitbulk this is ready for your initial review. I'll give it another pass tomorrow but the core changes and tests are all done (and working 🎉 )

johnnyshields:
@pitbulk this is ready for your review and merge. (Please review https://github.com/SAML-Toolkits/ruby-saml/pull/711 first)

johnnyshields:
@pitbulk FYI I am now running this branch in prod without issue.

pitbulk:
@johnnyshields, merged this one but rubocop needs to be fixed

PR #706, opened 2024-07-10
Remove stray "include REXML" + fix REXML deprecation

johnnyshields:
Also change deprecated REXML::Document.entity_expansion_limit to REXML::Security.entity_expansion_limit


pitbulk:
For reference:
https://www.rubydoc.info/stdlib/rexml/REXML%2FDocument.entity_expansion_limit

PR #710, opened 2024-07-10
[READY] Merge master into v2.x (please don't squash commits!)

johnnyshields:
Please merge this PR without squashing commits so the diff between v2.x and master will be fully caught-up. (This was [previously merged here](https://github.com/SAML-Toolkits/ruby-saml/pull/696) but it was done via squash)

https://github.com/SAML-Toolkits/ruby-saml/compare/v2.x...master


johnnyshields:
Replaced by https://github.com/SAML-Toolkits/ruby-saml/pull/725

PR #711, opened 2024-07-11
[READY] v2.0 - Consistently format cert/private key PEMs

johnnyshields:
Fixes https://github.com/SAML-Toolkits/ruby-saml/issues/636

I've tried to make this as compatible as possible with the legacy behavior while still handling the case stripping out text before/after the PEM string.

**All existing test cases behave as before** (* see [comment below](https://github.com/SAML-Toolkits/ruby-saml/pull/711#discussion_r1673990011)--there's one case that needed a minor alteration as the case itself was written incorrectly.)


lib/ruby_saml/utils.rb
```diff
@@ -84,77 +87,44 @@ def self.parse_duration(duration, timestamp=Time.now.utc)
...
-        cert = cert.gsub(/\n/, "")
-        cert = cert.gsub(/\s/, "")
-        cert = cert.scan(/.{1,64}/)
-        cert = cert.join("\n")
-        "-----BEGIN CERTIFICATE-----\n#{cert}\n-----END CERTIFICATE-----"
-      end
+    # @param cert [String] The original certificate(s)
+    # @param multi [true|false] Whether to return multiple keys delimited by newline
+    # @return [String|nil] The formatted certificate(s)
+    def format_cert(cert, multi: true)
```
johnnyshields, line 96: Here `multi: true` because that was the legacy behavior. It doesn't really make sense tho...

lib/ruby_saml/utils.rb
```diff
@@ -1,13 +1,16 @@
...
+require 'openssl'
+require 'ruby_saml/pem_formatter'
 
 module RubySaml
 
   # SAML2 Auxiliary class
   #
-  class Utils
+  module Utils
+    extend self
```
johnnyshields, line 12: Converted Utils to module

test/response_test.rb
```diff
@@ -1461,14 +1461,19 @@ def generate_audience_error(expected, actual)
         end
       end
 
-      it 'raise if an encrypted assertion is found and the sp private key is wrong' do
+      it 'raise if an encrypted assertion is found and the SP private key does not match cert' do
         settings.certificate = ruby_saml_cert_text
-        wrong_private_key = ruby_saml_key_text.sub!('A', 'B')
+        wrong_private_key = ruby_saml_key_text.sub!('Z', 'X')
```
johnnyshields, line 1466: This spec was kind of silly before, previously `ruby_saml_key_text.sub!('A', 'B')` was gsubbing `-----RSA PRIVATE KEY-----` to `RSB`, and not actually affecting the pkey body. I've improved it.

test/utils_test.rb
```diff
@@ -44,8 +44,8 @@ def result(duration, reference = 0)
     let(:formatted_chained_certificate) {read_certificate("formatted_chained_certificate")}
 
     it "returns empty string when the cert is an empty string" do
-      cert = ""
-      assert_equal "", RubySaml::Utils.format_cert(cert)
+      cert = ''
+      assert_nil RubySaml::Utils.format_cert(cert)
```
johnnyshields, line 48: The case of nil/empty/blank string now consistently returns nil.

lib/ruby_saml/pem_formatter.rb
```diff
@@ -0,0 +1,68 @@
...
+      algo = key.match(/((?:RSA|ECDSA|EC|DSA) )PRIVATE KEY/)&.[](1)
+      label = "#{algo}PRIVATE KEY"
+      format_pem(key, label)
+    end
+
+    # Strips all whitespace and the BEGIN/END lines,
+    # then splits the string into 64-character lines,
+    # and re-applies BEGIN/END labels
+    def format_pem(str, label)
+      str = str.gsub(/\s|-{5}\s*(BEGIN|END) [A-Z\d\s]+-{5}/, '').scan(/.{1,64}/).join("\n")
```
johnnyshields, line 64: This regex matches `\s` OR the header. Note `\s` is any whitespace, including `\n` and `\r`.

johnnyshields:
@pitbulk this is ready for your review and merge.

PR #716, opened 2024-07-22
Raise meaningful error when IDP descriptor cant be found

eriklovmo:
Here, raise a more meaningful error when the identity provider metadata does not contain an IDP descriptor, or no IDP descriptor matching the value of the optional `entity_id` option.

Using namespaced, specific errors makes the API easier and safer for user code.


johnnyshields:
Rather than doing this as a one-off, it would be better to look at _all errors_ RubySaml generates and make them into proper classes.

PR #718, opened 2024-08-27
add supports params that can be used for re-authentication

yviscuit:
Hi maintainers,

I want to verify whether authentication was performed when the SP requested force authentication(`ForceAuthn=true`).
This can be achived by obtaining the authentication time and method of the verification.
So, added getter methods for `AuthnInstant` and `AuthnContextClassRef` to `OneLogin::RubySaml::Response` class.

Please consider this suggestion.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -198,6 +198,27 @@ module OneLogin
         end
       end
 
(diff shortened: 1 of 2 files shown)


pitbulk (approved):

yviscuit:
Thanks!

PR #719, opened 2024-08-28
Update for frozen string compatibility

HParker:
This marks mutable strings to allow tests to pass with

    RUBYOPT="--enable-frozen-string-literal"

enabled.

This will be the default in future Ruby versions, so being compatible now is ideal.


johnnyshields:
This is not needed. v2.x already takes care of this with magic comments.

Issue #723, opened 2024-09-23
undefined method 'empty?' for an instance of OpenSSL::X509::Certificate with version 1.17.0


tobiasamft:
Hi, I'm observing the following error after updating ruby-saml from 1.16.0 to 1.17.0:
```
/usr/local/bundle/gems/ruby-saml-1.17.0/lib/onelogin/ruby-saml/settings.rb:316:in `validate_sp_certs_params!': undefined method `empty?' for an instance of OpenSSL::X509::Certificate (NoMethodError)

        cert     = certificate     && !certificate.empty?
                                                  ^^^^^^^
	from /usr/local/bundle/gems/ruby-saml-1.17.0/lib/onelogin/ruby-saml/settings.rb:309:in `get_all_sp_certs'
	from /usr/local/bundle/gems/ruby-saml-1.17.0/lib/onelogin/ruby-saml/settings.rb:213:in `get_sp_certs'
	from /usr/local/bundle/gems/ruby-saml-1.17.0/lib/onelogin/ruby-saml/settings.rb:252:in `get_sp_decryption_keys'
	from /usr/local/bundle/gems/ruby-saml-1.17.0/lib/onelogin/ruby-saml/response.rb:972:in `generate_decrypted_document'
	from /usr/local/bundle/gems/ruby-saml-1.17.0/lib/onelogin/ruby-saml/response.rb:70:in `initialize'
	from saml_response.rb:8:in `new'
	from saml_response.rb:8:in `<main>'
```

I've tested the following code snippet:
```
require 'onelogin/ruby-saml'

certificate = OpenSSL::X509::Certificate.new(File.read('ruby-saml.crt'))
private_key = File.read('ruby-saml.key')
saml_response = Base64.decode64(File.read('signed_message_encrypted_signed_assertion.xml.base64'))

response =
  OneLogin::RubySaml::Response.new(
    saml_response,
    settings: OneLogin::RubySaml::Settings.new(certificate:, private_key:)
  )
pp response
```

I've used the certificate, key, and xml from this repo to test it:
```
wget https://raw.githubusercontent.com/SAML-Toolkits/ruby-saml/refs/heads/master/test/certificates/ruby-saml.crt
wget https://raw.githubusercontent.com/SAML-Toolkits/ruby-saml/refs/heads/master/test/certificates/ruby-saml.key
[...]

PR #726, opened 2024-10-03
Allow SP certificates to be OpenSSL::X509::Certificate

tobiasamft:
This allows settings to accept instances of OpenSSL::X509::Certificate as service provider (SP) certificates.

Solves #723 

Version 1.16.0 was, at least partially, able to handle `OpenSSL::X509::Certificate` as input for settings.certificate (e.g. when using   `OneLogin::RubySaml::Response`).

Since `settings.get_sp_certs` is the only interface that is used to access certificates, it should be enough to test that interface with instances of `OpenSSL::X509::Certificate`. There are 3 ways to insert certs, all of them have been tested:
- settings.certificate
- settings.certificate_new
- settings.sp_cert_multi

Note that both deprecated interfaces `settings.get_sp_cert` and `settings.get_sp_cert_new` use `settings.get_sp_certs` internally. Thus, they are covered as well.

Same approach could be used for SP private key to accept `OpenSSL::PKey`.
Maybe it's a good idea to make all certificates from settings `attr_writer` for public and `attr_accessor` for private access to ensure that certs are accessed via `settings.get_sp_certs` only (but that would break current interface).

### DIFF
--- a/lib/ruby_saml/settings.rb
+++ b/lib/ruby_saml/settings.rb
@@ -375,13 +375,13 @@ module RubySaml
     def validate_sp_certs_params!
       has_multi = sp_cert_multi && !sp_cert_multi.empty?
       has_pk = private_key && !private_key.empty?
-      if has_multi && (has_cert?(certificate) || has_cert?(certificate_new) || has_pk)
+      if has_multi && (cert?(certificate) || cert?(certificate_new) || has_pk)
         raise ArgumentError.new("Cannot specify both sp_cert_multi and certificate, certificate_new, private_key parameters")
       end
     end
 
     # Check if a certificate is present.
-    def has_cert?(cert)
+    def cert?(cert)
       return true if cert.is_a?(OpenSSL::X509::Certificate)
 
       cert && !cert.empty?
--- a/lib/ruby_saml/settings.rb
+++ b/lib/ruby_saml/settings.rb
@@ -374,10 +374,14 @@ module RubySaml
     # Validate certificate, certificate_new, private_key, and sp_cert_multi params.
     def validate_sp_certs_params!
       has_multi = sp_cert_multi && !sp_cert_multi.empty?
-      has_pk = private_key && !private_key.empty?
-      if has_multi && (cert?(certificate) || cert?(certificate_new) || has_pk)
+      if has_multi && (cert?(certificate) || cert?(certificate_new) || pk?)
         raise ArgumentError.new("Cannot specify both sp_cert_multi and certificate, certificate_new, private_key parameters")
       end
+    end
+
+    # Check if private key exists and is not empty
+    def pk?
+      private_key && !private_key.empty?
     end
 
     # Check if a certificate is present.
@@ -392,14 +396,14 @@ module RubySaml
       certs = { :signing => [], :encryption => [] }
 
       sp_key = RubySaml::Utils.build_private_key_object(private_key)
-      cert = build_cert_object(certificate)
+      cert = RubySaml::Utils.build_cert_object(certificate)
       if cert || sp_key
         ary = [cert, sp_key].freeze
         certs[:signing] << ary
         certs[:encryption] << ary
       end
 
-      cert_new = build_cert_object(certificate_new)
+      cert_new = RubySaml::Utils.build_cert_object(certificate_new)
       if cert_new
         ary = [cert_new, sp_key].freeze
         certs[:signing] << ary
@@ -434,7 +438,7 @@ module RubySaml
           end
 
           certs[type] << [
-            build_cert_object(cert),
+            RubySaml::Utils.build_cert_object(cert),
             RubySaml::Utils.build_private_key_object(key)
           ].freeze
         end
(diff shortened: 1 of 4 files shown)


johnnyshields:
@tobiasamft can you check if this is solved on the v2.x branch? I think it might be already. If it is, we can close this PR b/c we are releasing v2.x soon.

tobiasamft:
@johnnyshields unfortunately v2.x does not solve this. Using `OpenSSL::X509::Certificate` as SP certificate still crashes with the following:
`git/ruby-saml/lib/ruby_saml/settings.rb:377:in 'validate_sp_certs_params!': undefined method 'empty?' for an instance of OpenSSL::X509::Certificate (NoMethodError)`

johnnyshields:
ok. Can you raise the PR to the v2.x branch then please? I will review it.

lib/onelogin/ruby-saml/settings.rb
```diff
@@ -377,6 +382,12 @@ def get_sp_certs_multi
         certs.each { |_, ary| ary.freeze }
         certs
       end
+
+      def build_cert_object(cert)
+        return cert if cert.is_a?(OpenSSL::X509::Certificate)
+
+        OneLogin::RubySaml::Utils.build_cert_object(cert)
```
pitbulk, line 389: Why not adapt directly the Utils.build_cert_object to be able to handle string and OpenSSL::X509::Certificate?

tobiasamft: Yes, I was thinking about that, too. But then I decided to catch the problem as early as possible. And if it's fixed in Utils, it creates kind of a dependency. Utils would fix problems that come from Settings. Settings would not correctly work on it's own.

lib/ruby_saml/settings.rb
```diff
@@ -438,5 +443,11 @@ def get_sp_certs_multi
       certs.each { |_, ary| ary.freeze }
       certs
     end
+
+    def build_cert_object(cert)
+      return cert if cert.is_a?(OpenSSL::X509::Certificate)
+
+      OneLogin::RubySaml::Utils.build_cert_object(cert)
```
johnnyshields, line 450: let's just change this method in OneLogin::RubySaml::Utils to return a cert if one is given as an argument (it should be "idempotent")

johnnyshields:
@tobiasamft see comment

lib/ruby_saml/settings.rb
tobiasamft: 👍 Your code, your decision 😄 Will change that.

lib/ruby_saml/settings.rb
```diff
@@ -373,15 +373,20 @@ def get_all_sp_certs
...
         raise ArgumentError.new("Cannot specify both sp_cert_multi and certificate, certificate_new, private_key parameters")
       end
     end
 
+    # Check if a certificate is present.
+    def cert?(cert)
+      return true if cert.is_a?(OpenSSL::X509::Certificate)
+
+      cert && !cert.empty?
+    end
```
johnnyshields, line 388: all changes in this file can be rolled back

tobiasamft: Without it we get the initial NoMethodError again:
```
NoMethodError: undefined method `empty?' for an instance of OpenSSL::X509::Certificate
    lib/ruby_saml/settings.rb:377:in `validate_sp_certs_params!'
```

johnnyshields: OK I will check this later, thanks!

pitbulk: In what status is this?

johnnyshields: @pitbulk this change is fine. @tobiasamft how about also adding a `pk?` method which does the same thing instead of has_pk?

tobiasamft: Hi @johnnyshields, I've added `pk?` now. How about `has_multi` should it be treated the same way?
And how about `lib/ruby_saml/utils.rb`, rubocop complains about module size (I've just added 2 lines)?

lib/ruby_saml/settings.rb
```diff
@@ -373,15 +373,24 @@ def get_all_sp_certs
...
+      has_multi = sp_cert_multi && !sp_cert_multi.empty?
+      if has_multi && (cert?(certificate) || cert?(certificate_new) || pk?)
         raise ArgumentError.new("Cannot specify both sp_cert_multi and certificate, certificate_new, private_key parameters")
       end
     end
 
+    # Check if private key exists and is not empty
+    def pk?
+      private_key && !private_key.empty?
+    end
```
johnnyshields, line 385: Sorry I meant to add:

```
      return true if cert private_key.is_a?(OpenSSL::PKey::PKey)
```

lib/ruby_saml/settings.rb
johnnyshields: See my comment above about allowing OpenSSL::PKey::PKey. Please add tests for it then I'll handle the rest on this one, including rubocop

johnnyshields:
@tobiasamft thanks for all your help on this, I've reworked and committed this here: https://github.com/SAML-Toolkits/ruby-saml/pull/732

PR #727, opened 2024-10-22
Introduce thread safety to SAML schema read

e-ikuta:
## Problem
As the related PR below, use of Dir.chdir is not thread-safe. This usage was causing exceptions to be raised when running on Puma and in other multi-threaded environments.

Related:
- https://github.com/SAML-Toolkits/ruby-saml/pull/175

## How to fix
This patch replaces `Dir.chdir` with thread-safe `File.open` method.

## Additional information
This patch also removes `@@mutex`, which became unnecessary.


alpaca-tc:
@johnnyshields @pitbulk  If everything looks good to you, I would greatly appreciate it if you could consider merging this PR 😄 

In the service I manage, we occasionally encounter a rare issue where file reading fails with an error in production. After investigating, we identified that the root cause is a `Dir.chdir` call in another thread, which changes the current working directory unexpectedly.

This PR would resolve the issue and ensure stable file reading.

スクリーンショット_2025-01-08_18_42_39-2

johnnyshields:
@pitbulk lgtm, please merge.

PR #728, opened 2024-10-23
Support frozen strings with REXML

frederikspang:
Tests are passing with `export RUBYOPT=--enable-frozen-string-literal`

### DIFF
--- a/lib/onelogin/ruby-saml/utils.rb
+++ b/lib/onelogin/ruby-saml/utils.rb
@@ -32,7 +32,9 @@ module OneLogin
           (\d+)W                    # 8: Weeks
         )
       $)x.freeze
-      UUID_PREFIX = +'_'
+
+      UUID_PREFIX = '_'
+      @@prefix = '_'.freeze
 
       # Checks if the x509 cert provided is expired.
       #
@@ -252,6 +254,8 @@ module OneLogin
       # @param status_message [Strig] StatusMessage value
       # @return [String] The status error message
       def self.status_error_msg(error_msg, raw_status_code = nil, status_message = nil)
+        error_msg = error_msg.dup
+
         unless raw_status_code.nil?
           if raw_status_code.include? "|"
             status_codes = raw_status_code.split(' | ')
@@ -262,11 +266,11 @@ module OneLogin
           else
             printable_code = raw_status_code.split(':').last
           end
-          error_msg += ', was ' + printable_code
+          error_msg << ', was ' + printable_code
         end
 
         unless status_message.nil?
-          error_msg += ' -> ' + status_message
+          error_msg << ' -> ' + status_message
         end
 
         error_msg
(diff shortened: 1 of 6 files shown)


lib/onelogin/ruby-saml/utils.rb
```diff
@@ -32,7 +32,7 @@ class Utils
           (\d+)W                    # 8: Weeks
         )
       $)x.freeze
-      UUID_PREFIX = '_'
+      UUID_PREFIX = +'_'
```
johnnyshields, line 35: Constants should remain frozen. Instead, call `UUID_PREFIX.dup` where it is used, if necessaty.

frederikspang: It's used with `.replace` which won't work. Reassigning the constant is "bad".
I'm replacing it with a variable, but should we keep it for compatibility, or remove it and bump a major version?

frederikspang:
@johnnyshields I've adjusted some stuff from review. UUID_PREFIX is used as a constant, and `.replace` with Utils.set_prefix, which is not working out great :). I've replaced it with a variable. Should we deprecate or remove the (now unused) constant?

lib/onelogin/ruby-saml/utils.rb
```diff
@@ -32,7 +32,9 @@ class Utils
           (\d+)W                    # 8: Weeks
         )
       $)x.freeze
+
       UUID_PREFIX = '_'
+      @@prefix = '_'.freeze
```
johnnyshields, line 37: not necessary to call .freeze here.

johnnyshields:
@frederikspang yikes wow we were calling replace on a constant 🤦‍♂️ . The way you have it now with `@@prefix` is fine.

lib/onelogin/ruby-saml/utils.rb
frederikspang: @johnnyshields Fixed

johnnyshields:
@pitbulk frozen string stuff is handled already on 2.x. This is a "backport" in case we want it on 1.x branch, but the alternative here would be to just close this PR and tell users that you must use 2.x for Ruby 3.4+. Which is reasonable in my opinion.

johnnyshields:
Mutable constant addressed here: https://github.com/SAML-Toolkits/ruby-saml/pull/735

frederikspang:
@johnnyshields Frozen strings is not exclusive for Ruby 3.4 - It's just moving towards default frozen, with "chilled string". 

I think that not supporting frozen strings for < 3.4 would be a mistake, when the backport is somewhat simple for 1.x branch.

frederikspang:
Specs may have been solved for older ruby versions, with using .dup instead of `+""` for unfreezing. Could you rerun specs?

johnnyshields:
@frederikspang we are close to releasing v2.x branch so this will be a moot point...

johnnyshields:
It's @pitbulk 's call whether he wants to release for 1.x branch.

pitbulk:
@frederikspang, it seems for ruby >= 3.4.0 we need to add mutex_m gem on the gemspec, can you add it?

frederikspang:
@pitbulk Got it :)

gstokkink:
@pitbulk can we get a new release for this? 😄

PR #730, opened 2024-12-11
Implement handling of CBC padding for symmetric decryption.

vcsjones:
AES and 3DES, when used with encrypted assertions with CBC padding, use a form of padding called ISO 10126 and is documented here: https://www.w3.org/TR/xmlenc-core1/#sec-Padding.

The current implementation calls this "noise" and uses a regular expression to attempt to extract the plaintext without the padding. Rather than use a regular expression, this implements correct handling of the padding.]]

This is more of a correctness thing than addressing any particular issue. Since the padding is random it could, in theory, be something like `NameID>` which would incorrectly be handled by the regex.

The unit tests already handle the padding aspect. Removing the regular expressions without the padding handling caused test failures. Rather than introduce any new tests, existing tests are used to validate the changes.

### DIFF
--- a/lib/onelogin/ruby-saml/response.rb
+++ b/lib/onelogin/ruby-saml/response.rb
@@ -1029,7 +1029,7 @@ module OneLogin
       # @return [REXML::Document] The decrypted EncryptedAssertion element
       #
       def decrypt_assertion(encrypted_assertion_node)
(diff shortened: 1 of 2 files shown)


vcsjones:
Closing this out for now, I think there are some opportunities to improve testing here.

PR #733, opened 2025-01-11
[READY] v2.x - Allow `RubySaml::Utils.is_cert_expired` and `is_cert_active` to accept an optional time argument

johnnyshields:


lib/ruby_saml/utils.rb
```diff
@@ -35,32 +35,30 @@ module Utils
...
-    # @param timestamp [Integer] The unix timestamp we should apply the
-    #                            duration to. Optional, default to the
-    #                            current time.
-    #
+    # @param timestamp [Time|Integer] The unix timestamp we should apply the
+    #   duration to. Optional, default to the current time.
     # @return [Integer] The new timestamp, after the duration is applied.
-    #
-    def parse_duration(duration, timestamp=Time.now.utc)
+    def parse_duration(duration, timestamp = Time.now)
```
johnnyshields, line 61: Safe to do `Time.now` without `.utc` here because it's converted to UTC below.

PR #735, opened 2025-01-11
[READY] v2.x - Rework SP request UUID generation to remove mutable constant

johnnyshields:
This PR reworks the way SP request UUID generation is handled.
- Replace the mutable `RubySaml::Utils::UUID_PREFIX` constant and `set_prefix` method with `Settings.sp_uuid_prefix`
   - (Mutable constants are not thread-safe, violate principle of least surprise, etc.)
- Make Authnrequest, etc. `uuid` attribute to be immutable.
- Initialize Authnrequest, etc. `uuid` attribute when `create` is called, based on `Settings.sp_uuid_prefix`, not when instantiating the Ruby object.

image


johnnyshields:
@pitbulk would like to get your feedback on this. I am OK to keep the `Authnrequest.uuid =` mutator method if there's really a use case for it.

pitbulk:
@johnnyshields it seems tests are failing.

johnnyshields:
@pitbulk This is ready for final review.

johnnyshields:
1 failure die to that JRuby Zlib issue still. Seems my previous fix didn't resolve it...

pitbulk:
@johnnyshields, now that prefix can be defined, maybe should be mention what the spec requires in terms of the possible values set to the ID attribute:

> The type xsd:ID is used for an attribute that uniquely identifies an element in an XML document. An xsd:ID value must be an NCName. This means that it must start with a letter or underscore, and can only contain letters, digits, underscores, hyphens, and periods.

johnnyshields:
@pitbulk please check now.

PR #736, opened 2025-01-11
[WIP] v3.0 - Replace REXML with Nokogiri

johnnyshields:
**UPDATE: This PR is abandoned, but please leave it open. I will close it when Nokogiri migration work is completed.**

---

Fixes https://github.com/SAML-Toolkits/ruby-saml/issues/729
Fixes https://github.com/SAML-Toolkits/ruby-saml/issues/707

Currently Ruby Saml uses a mix of REXML and Nokogiri. This is not ideal. We should use Nokogiri everywhere, because it can do everything REXML can do.

In addition, note:
- Ruby Saml is currently setting `REXML::Security.entity_expansion_limit = 0` which affects it for all gems in the project.
- The above is not thread-safe, while Nokogiri is thread-safe by design.
- REXML [does not use Semver](https://github.com/ruby/rexml/issues/211#issuecomment-2398870554), and [exposes users to sudden breakage](https://github.com/SAML-Toolkits/ruby-saml/issues/729).

This refactor will probably get rid of `RubySaml::XML::Document`, etc.. Currently these are subclasses of `REXML::Document`. What we should do instead is make them procedural classes like `DocumentSigner` and `SignedDocumentValidator`. Hence I'm thinking it will be v3.0


pitbulk:
@johnnyshields, let's replace REXML with Nokogiri on V2 and in V3 in addition, we can let's get rid of get rid of RubySaml::XML::Document as you suggested.

pitbulk:
@johnnyshields , any progress here? we need to drop REXML before the v2 release and I want to also plan to do that on master branch.

johnnyshields:
@pitbulk it's way too big a change, and potentially could introduce issues. Let's aim to do it in 3.0 and release what we have as 2.0

kwent:
Looking forward to this.

johnnyshields:
Abandoned this branch, see my other PRs

PR #748, opened 2025-03-11
[READY] V2.x - Nokogiri Upgrade Part 3 - Shim XMLSecurity so it raises deprecation warnings and errors

johnnyshields:
Previously for 2.x I was trying to make XMLSecurity backwards compatible. With the upcoming Nokogiri refactor, this is more difficult to do than I realized.

A better approach will be to simply shim the old XMLSecurity classes so they raise clear deprecations and NoMethodErrors if any one tries to use them--assuming they have any basic tests in their app, the tests will fail. Since XMLSecurity functionality is not the main focus of RubySaml anyway, this should be safe to do.

It may be possible to restore functionality to some of these methods after the Nokogiri migration is complete. We'll have to see.


johnnyshields:
@pitbulk This is ready for merge.

johnnyshields:
@pitbulk this is ready. Please merge in #754 first, then this one.

PR #759, opened 2025-03-16
[WIP] v2.x - Nokogiri Refactor Part X - Conversion of SignedDocument (attempt to SignedDocumentInfo class) 

johnnyshields:


p-:
@pitbulk Yes I will help 👍  But I think I can't start before next week.

johnnyshields:
Hi, yes I agree I am intending to make the code more functional. This PR was intended firstly to get all tests green while changing relatively little.

I may not have time to finish for a few weeks however.

johnnyshields:
@ahacker1-securesaml I agree with all those points, that sounds along the right lines. I will need more time to get my head around it, the code is really spaghetti today, and the real trick is getting all the tests to pass (fortunately the test suite is quite robust!!)

johnnyshields:
@pitbulk making the signatures validatable in the tests would be a huge help to give us flexibility in refactoring.

taylorreis:
@ahacker1-securesaml, would you be able to add some context to the following?

> ResponseValidator -> change to Assertion Validator, we don't care about the Response, since IdPs don't even sign it.

Signing the `Response` alone is something that identity providers can do in practice (Entra ID is an example). The `Assertion` should inherit the signature.

taylorreis:
Agree with the principles described, but not supporting a configuration that signs the `Response` alone would be a breaking change, given that it's currently supported in the wild:
- [Entra](https://learn.microsoft.com/en-us/entra/identity/enterprise-apps/certificate-signing-options#certificate-signing-options)
- [Key Cloak](https://docs.anchore.com/3.0/docs/overview/sso/examples/keycloak/keycloak_sso_client_settings1.png)

johnnyshields:
It would definitely help us to get a library of all the vendor IDP SAML variations in the wild... maybe some other SAML lib already has one?

As it stands the current RubySaml test suite is pretty good, it should protect us from a pretty wide range of possible regressions.

pitbulk:
In terms of security, I agree that if the Response is not signed, and the StatusCode is not "protected" anyone can modify it, but in terms of business, if I receive a StatusCode != Success, I don't need to do anything else, I will simply reject this SAMLResponse and don't grant access to the app, the same applies to malformed XMLs received or XMLs wellformed, but that we want to not allow (multiple Assertions, more than 2 Signatures).

dblessing:
@johnnyshields This PR works fine with omniauth-saml. Anything specific I can help test?

pitbulk:
Understood, thanks for your hard work!

johnnyshields:
@pitbulk I think we should merge this branch into v2.x regardless. I realize there are more refactors needed, but perhaps we can track those in a follow-up ticket?

pitbulk:
@johnnyshields, I was able to reorder the validations and adjust tests and payloads:
https://github.com/SAML-Toolkits/ruby-saml/pull/764

pitbulk:
Sorry for the delay!, I'm doing some final test this weekend and gonna be released next week.

UPGRADING.md
```diff
@@ -74,24 +74,33 @@ settings.security[:signature_method] = RubySaml::XML::RSA_SHA1
 ### Replacement of REXML with Nokogiri
 
 RubySaml `1.x` used a combination of REXML and Nokogiri for XML parsing and generation.
-In `2.0.0`, REXML has been replaced with Nokogiri. This change should be transparent
-to most users, however, see note about Custom Metadata Fields below.
+In `2.0.0`, REXML has been replaced with Nokogiri. As a result, there are minor differences
+in how XML is generated, ncluding SAML requests and SP Metadata:
```
pitbulk: **i**ncluding

PR #785, opened 2026-02-01
Use cgi/escape for Ruby 4.0 compatibility

joshwetzel:
Ruby 4.0 removed the full CGI library from the standard library, which produces a warning. Since the code only uses `CGI.escape`, we can require `cgi/escape` instead of the full `cgi` library. This provides just the escape/unescape functionality and is compatible with both Ruby 3.x and 4.x.

```
warning: CGI library is removed from Ruby 4.0. Please use cgi/escape instead for
CGI.escape and CGI.unescape features
```


joshwetzel:
@pitbulk thanks for merging https://github.com/SAML-Toolkits/ruby-saml/pull/784! Here's another one for you if you haven't seen it already.

johnnyshields:
@pitbulk can we release RubySaml 2.0 -- I already fixed this there, along with many other bugs.

PR #790, opened 2026-05-03
Raise `ArgumentError` when `nil settings` passed to some methods

alstrocrack:
## Status
**READY**

## Migrations
NO

## Description

Several methods across `OneLogin::RubySaml::Authrequest`, `Logoutrequest`, `SloLogoutresponse`, and `Metadata` accepted `nil` as their `settings` argument according to YARD comments (`[OneLogin::RubySaml::Settings|nil]`), but in practice passing `nil` causes a `NoMethodError` immediately when the method tries to call methods on `settings`.

This PR:

- Adds an explicit `ArgumentError` guard at the top of each affected public method, providing a clear error message instead of a cryptic `NoMethodError`
- Corrects the YARD `@param` annotations to remove `|nil`, accurately reflecting that `nil` is not a valid value
- Removes the dead-code private method `encode_raw_saml` from `SamlMessage`

`ArgumentError` was chosen over the library's custom `SettingError` because passing `nil` is a caller-side programming error (wrong argument), whereas `SettingError` is reserved for cases where a settings object exists but contains invalid or missing configuration values.

## Related PRs

N/A

## Todos
- [x] Tests
- [x] Documentation

## Deploy Notes

N/A

## Steps to Test or Reproduce

N/A

- `Authrequest#create`, `#create_params`, `#create_authentication_xml_doc`
- `Logoutrequest#create`, `#create_params`, `#create_logout_request_xml_doc`
- `SloLogoutresponse#create`, `#create_params`, `#create_logout_response_xml_doc`
- `Metadata#generate`

### DIFF
--- a/lib/onelogin/ruby-saml/saml_message.rb
+++ b/lib/onelogin/ruby-saml/saml_message.rb
@@ -100,17 +100,6 @@ module OneLogin
         saml
       end
 
-      # Deflate, base64 encode and url-encode a SAML Message (To be used in the HTTP-redirect binding)
-      # @param saml [String] The plain SAML Message
-      # @param settings [OneLogin::RubySaml::Settings|nil] Toolkit settings
-      # @return [String] The deflated and encoded SAML Message (encoded if the compression is requested)
-      #
-      def encode_raw_saml(saml, settings)
-        saml = deflate(saml) if settings.compress_request
-
-        CGI.escape(encode(saml))
-      end
-
       # Base 64 decode method
       # @param string [String] The string message
       # @return [String] The decoded string
--- a/lib/onelogin/ruby-saml/authrequest.rb
+++ b/lib/onelogin/ruby-saml/authrequest.rb
@@ -95,7 +95,7 @@ module OneLogin
       end
 
       # Creates the SAMLRequest String.
-      # @param settings [OneLogin::RubySaml::Settings|nil] Toolkit settings
+      # @param settings [OneLogin::RubySaml::Settings] Toolkit settings
       # @return [String] The SAMLRequest String.
       #
       def create_authentication_xml_doc(settings)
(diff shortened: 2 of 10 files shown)


pitbulk:
@alstrocrack, can you extend the review to other classes? Please add guardrails as well to avoid the NoMethodError errors, thanks!

lib/onelogin/ruby-saml/saml_message.rb
```diff
@@ -100,17 +100,6 @@ def decode_raw_saml(saml, settings = nil)
         saml
       end
 
-      # Deflate, base64 encode and url-encode a SAML Message (To be used in the HTTP-redirect binding)
-      # @param saml [String] The plain SAML Message
-      # @param settings [OneLogin::RubySaml::Settings|nil] Toolkit settings
-      # @return [String] The deflated and encoded SAML Message (encoded if the compression is requested)
-      #
-      def encode_raw_saml(saml, settings)
```
alstrocrack, line 108: I found this dead code. What do you think about me deleting this code because I thought it was related to these changes? As far as I checked, this code isn’t called from anywhere and is a private method.

alstrocrack:
@pitbulk 

I added some changes to this PR. I searched for incorrect comments regarding `nil settings` across this repo and fixed all of them.

I thought `ArgumentError` was more appropriate than `SettingsError`, so I used `ArgumentError`. However, I initially suggested using `SettingsError` as a `nil` guard error.

I also found some other places where similar `nil` guard fixes could be applied. However, since this PR was originally intended only to fix comments, I limited the changes to the sources related to those comments for now.

Could you check these changes again?

alstrocrack:
@pitbulk 

Just a reminder, in case you might have overlooked this.

